PR #7, opened 2014-03-25
XCOMMONS-553 : Releasing fails on windows os

sbaloul:


tmortagne:
Closing since there is no reason to apply this anymore.

PR #13, opened 2016-03-23
extension-repository-http: implemented HTTP proxy authentification.

nE0sIghT:
HttpClientBuilder.useSystemProperties method [do not takes into account](http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/client/HttpClientBuilder.html) login and password for authentification against proxy.
This change allows to use http.proxyUser and http.proxyPassword JVM system properties to authentificate against proxy server.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-repositories/xwiki-commons-extension-repository-http/src/main/java/org/xwiki/extension/repository/http/internal/DefaultHttpClientFactory.java
+++ b/xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-repositories/xwiki-commons-extension-repository-http/src/main/java/org/xwiki/extension/repository/http/internal/DefaultHttpClientFactory.java
@@ -22,6 +22,7 @@ package org.xwiki.extension.repository.http.internal;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+import org.apache.http.HttpHost;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.CredentialsProvider;
@@ -57,6 +58,35 @@ public class DefaultHttpClientFactory implements HttpClientFactory
         // Pre-configure with everything configured at JVM level
         httpClientBuilder.useSystemProperties();
 
+        HttpHost proxy = null;
+        CredentialsProvider proxyCredentialsProvider = null;
+        String proxyHost = System.getProperty("http.proxyHost");
+        String proxyPort = System.getProperty("http.proxyPort");
+        if ((proxyHost != null) && !proxyHost.isEmpty()) {
+            int port = 3128;
+            if (proxyPort != null && !proxyPort.isEmpty()) {
+                port = Integer.parseInt(proxyPort);
+            }
+
+            proxy =  new HttpHost(proxyHost, port);
+            httpClientBuilder.setProxy(proxy);
+        }
+
+        if (proxy != null) {
+            String proxyUser = System.getProperty("http.proxyUser");
+            if (proxyUser != null && !proxyUser.isEmpty()) {
+                String proxyPassword = System.getProperty("http.proxyPassword");
+
+                proxyCredentialsProvider = new BasicCredentialsProvider();
+                proxyCredentialsProvider.setCredentials(
+                        new AuthScope(proxy),
+                        new UsernamePasswordCredentials(proxyUser, proxyPassword)
+                );
+
+                httpClientBuilder.setDefaultCredentialsProvider(proxyCredentialsProvider);
+            }
+        }
+
         // Setup user agent
         httpClientBuilder.setUserAgent(this.configuration.getUserAgent());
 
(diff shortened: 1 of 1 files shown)


tmortagne:
This looks like a bug on HTTPClient side to me.

vmassol:
Why do you say a bug thomas, the doc at http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/client/HttpClientBuilder.html clearly lists what it handles and `http.proxyPassword` and  `http.proxyUser` are not listed. Could be an improvement (although they may have not supported them voluntarily by default) but I don't see why it would be a bug.

vmassol:
Also, even if it were a bug, we still need to work around it.

tmortagne:
I say a bug because it does not make much sense for useSystemProperties() to not handled something as standard as http.proxyUser and http.proxyPassword. IMO they were just forgotten and should be reported, if it was on purpose there would be a note about those two properties in the documentation.

tmortagne:
Thanks for the contribution @nE0sIghT by the way but I would like to understand why HTTPClient does not support this in the first place before reimplementing on our side the proxy. Also note that this workaround only works for HTTP and not HTTPS proxy from what I understand (at least the standard properties for https are supposed to be https.proxyUser and https.proxyPassword)

nE0sIghT:
To say the truth http.proxy(User|Password) never was some standard as I know.
[Java documentation](https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html) only says about http.proxyHost, http.proxyPort and http.nonProxyHosts.
For an example, Liferay uses com.liferay.portal.util.HttpImpl.proxy.(username|password).

I do not think this must be fixed in commons HTTP client.

tmortagne:
> Java documentation only says about http.proxyHost, http.proxyPort and http.nonProxyHosts.

It also talk about https.proxyHost and https.proxyPort.

For proxyUser the closest from a standard documentation I could find is http://www.ibm.com/support/knowledgecenter/was_beta/com.ibm.websphere.base.doc/ae/twbs_configwbsclient2webproxy.html which is also talking about http.* vs https.*.

xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-repositories/xwiki-commons-extension-repository-http/src/main/java/org/xwiki/extension/repository/http/internal/DefaultHttpClientFactory.java
```diff
@@ -57,6 +58,35 @@ public CloseableHttpClient createClient(String user, String password)
...
+        String proxyHost = System.getProperty("http.proxyHost");
+        String proxyPort = System.getProperty("http.proxyPort");
+        if ((proxyHost != null) && !proxyHost.isEmpty()) {
+            int port = 3128;
+            if (proxyPort != null && !proxyPort.isEmpty()) {
+                port = Integer.parseInt(proxyPort);
+            }
+
+            proxy =  new HttpHost(proxyHost, port);
+            httpClientBuilder.setProxy(proxy);
```
tmortagne: Seems to me setting the right credential provider should be enough. Setting the proxy is supposed to be useSystemProperties() job, then it will ask the credential provider if it needs authentication.

nE0sIghT: Agreed. However we need to construct proxy HttpHost object to pass it to credential provider.
This part is not clear for me if auto detected system proxy used.
I think the proper way is to extract all auto detected system proxies (as commons HTTP client do it) and set credentials for each of them.

tmortagne: An much simpler alternative to all that might be to just set a default java.net.Authenticator which return whatever is indicated in http.proxyUser/Pass if called with RequestorType.PROXY.

nE0sIghT: Looks nice! I think this is a best method.

sdumitriu:
Check out https://github.com/phenotips/phenotips/tree/master/components/proxy-authentication

nE0sIghT:
@sdumitriu
As I see this implementation wil reveal proxy login/password to any website. At least getRequestorType method should be used.

sdumitriu:
@nE0sIghT True, but the example is how this could be accomplished globally, not just for extension-repository-http, and in a manner that doesn't add complexity. The authenticator itself is as simple as possible, and it should indeed check the authenticator type, and could be extended to take into account different http and https settings.

tmortagne:
Yep this is a good illustration of what I was talking about, plus it's already XWiki style :)

But this one is missing a few things IMO:
* it should wrap any pre existing default Authenticator (and put it back in the dispose() of the listener probably)
* it should return the login/password only if the type is PROXY, otherwise fallback on wrapped Authenticator
* it should also support https.proxyUser/Passwor IMO since I saw several mentions of this around (and they make sense anyway since that's how proxy host and ports are working too)

PR #14, opened 2016-05-22
[Misc] Fix violation for new property 'maxForVoid' for 'ReturnCountCheck'

rnveach:
This PR is being created because of a new violation found in CheckStyle's regression of XWiki.
Issue: https://github.com/checkstyle/checkstyle/issues/3143
PR: https://github.com/checkstyle/checkstyle/pull/3180#issuecomment-219337128

A new property for `ReturnCountCheck` was added, 'maxForVoid'. This property allows users to count `return;` statements separately from `return expression;` statements. The default value of this property is 1 and this caused the following similar violations in XWiki:
````[ERROR] [src/main/java/org/xwiki/job/internal/DefaultJobProgress.java:180,5](https://github.com/xwiki/xwiki-commons/blob/master/xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/DefaultJobProgress.java#L180-L206) (coding) ReturnCount: Return count is 2 (max allowed is 1).````

You can either accept this PR and not have an issue when you upgrade CS with this new feature in the future,
or make your own change/fix later when you do upgrade CS.
Please let us know which way you intend to go.

Feel free to ask any other questions or propose any additional changes.
Thanks.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/job/history/internal/ExtensionJobHistoryRecorder.java
+++ b/xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/job/history/internal/ExtensionJobHistoryRecorder.java
@@ -100,19 +100,16 @@ public class ExtensionJobHistoryRecorder extends AbstractEventListener
 
     private void onJobStarted(Job job)
     {
-        if (!(job.getRequest() instanceof ExtensionRequest)) {
-            // This is not an extension job.
-            return;
-        }
+        if (job.getRequest() instanceof ExtensionRequest) {
+            if (job.getStatus() instanceof AbstractJobStatus && isSubJob((AbstractJobStatus<?>) job.getStatus())) {
+                // We record only the jobs that have been triggered explicitly or that are part of a replay.
+                return;
+            }
 
-        if (job.getStatus() instanceof AbstractJobStatus && isSubJob((AbstractJobStatus<?>) job.getStatus())) {
-            // We record only the jobs that have been triggered explicitly or that are part of a replay.
-            return;
-        }
-
-        String jobId = StringUtils.join(job.getRequest().getId(), '/');
-        if (jobId != null) {
-            this.answers.put(jobId, new HashMap<String, QuestionRecorder<Object>>());
+            String jobId = StringUtils.join(job.getRequest().getId(), '/');
+            if (jobId != null) {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by sdumitriu** [comment]
Although the code changes aren't affecting behavior in any way, I still prefer an `if (not OK) return` style of code instead of wrapping everything in an `if (OK)` block. Wouldn't it be simpler to change the [checkstyle configuration](https://github.com/xwiki/xwiki-commons/blob/master/xwiki-commons-tools/xwiki-commons-tool-verification-resources/src/main/resources/checkstyle.xml) to allow more void returns?

**Comment by romani** [comment]
> Wouldn't it be simpler to change the checkstyle configuration to allow more void returns?

yes you can, it is a matter of style. We do not want to force you change smth if you do not like.
Your project already have 5 a limit
https://github.com/xwiki/xwiki-commons/blob/master/xwiki-commons-tools/xwiki-commons-tool-verification-resources/src/main/resources/checkstyle.xml#L303
```
    <module name="ReturnCount">
      <property name="max" value="5"/>
    </module>
```

> I still prefer an if (not OK) return style of code instead of wrapping everything in an if (OK) block. 

up to you. I was on the same position for a while. We even introduced https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/ReturnCountExtendedCheck.java to allow simliar cases.
But now I am become more strict in return count policy as it help more than do damage to code.
But this purely subjective approach.

@sdumitriu, Do you against all changes of this PR ?
if yes - just close it.
if no - let us know we could revert some of them and let you merge this PR. @rnveach  did good optimizations in some part of xwiki code, might be good to keep them.

@rnveach , looks like we need to make a fork from xwiki for now and keep our CI use forked code where we will make maxForVoid=5 to be equal to current policy of xwiki.
After xwiki update to checkstyle 6.19 version, we could remove our fork.

rnveach:
@sdumitriu What do you want to do with this PR?

tmortagne:
Thanks for the pull request !

I don't understand how I missed this PR. I recently upgraded to Checkstyle 6.19 and simply added

          <property name="maxForVoid" value="5"/>

> You can either accept this PR and not have an issue when you upgrade CS with this new feature in the future,

Actually there is other violations on xwiki-rendering and xwiki-platform and I was too lazy to have a debate about them just yet.

I will close this PR for now and we need to have a discussion about this on XWiki mailing if someone really feel it needs to be changed (I don't have a strong enough opinion on this either way to start this discussion).

PR #16, opened 2016-08-23
Update Revapi versions to include a fix for https://github.com/revapi/revapi/issues/39

metlos:


vmassol:
Thanks so much Lukas! :)

PR #19, opened 2017-08-02
[misc] updated Checkstyle version from 8.0 to 8.1 and tests in xwiki-common-tool-verification-resources to remove BaseCheckTestSupport and extend AbstractModuleTestSupport

subkrish:
This PR is being created because of the newer version of Checkstyle that was released -> `8.1` and also that `BaseCheckTestSupport` is now deprecated and will be removed in https://github.com/checkstyle/checkstyle/issues/4867 and by the time Checkstyle version `8.2` is released it will be removed.

All those tests that extend `BaseCheckTestSupport` will now extend `AbstractModuleTestSupport` which was added in Checkstyle version `8.1`. 

Please feel free to ask any other questions or propose any additional changes that you may require.
Thank You.

### DIFF
--- a/xwiki-commons-tools/xwiki-commons-tool-verification-resources/src/test/java/org/xwiki/tool/checkstyle/UnstableAnnotationCheckTest.java
+++ b/xwiki-commons-tools/xwiki-commons-tool-verification-resources/src/test/java/org/xwiki/tool/checkstyle/UnstableAnnotationCheckTest.java
@@ -19,13 +19,10 @@
  */
 package org.xwiki.tool.checkstyle;
 
-import java.io.File;
-import java.io.IOException;
-
 import org.junit.Before;
 import org.junit.Test;
 
-import com.puppycrawl.tools.checkstyle.BaseCheckTestSupport;
+import com.puppycrawl.tools.checkstyle.AbstractModuleTestSupport;
 import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
 import com.puppycrawl.tools.checkstyle.utils.CommonUtils;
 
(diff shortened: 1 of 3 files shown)


xwiki-commons-tools/xwiki-commons-tool-verification-resources/src/test/java/org/xwiki/tool/checkstyle/UnstableAnnotationCheckTest.java
```diff
@@ -35,15 +32,15 @@
...
+public class UnstableAnnotationCheckTest extends AbstractModuleTestSupport
 {
     private DefaultConfiguration checkConfig;
 
     @Before
     public void setUp()
     {
         this.checkConfig = createCheckConfig(UnstableAnnotationCheck.class);
-        this.checkConfig.addAttribute("currentVersion", "8.1-SNAPSHOT");
+        this.checkConfig.addAttribute("currentVersion", "8.2-SNAPSHOT");
```
mflorea: I don't think you need to change this version because it's not the Checkstyle version.

subkrish: @mflorea The only reason I changed it was because when the Checkstyle version was `8.0`, here `8.1-SNAPSHOT` was used for testing. I just thought it would be appropriate to make the change since I was updating the version to `8.1`.

rnveach: @mflorea Can you explain what this version is since it looks like it is so closely related to the Checkstyle version?

@subkrish Please verify the build and run doesn't fail if you revert these snapshot versions in tests. If so, please push the changes requested.

subkrish: @mflorea @rnveach Changes have been reverted from `8.2` back to `8.1`. Build does pass with original snapshot version in tests. Done.

mflorea:
Thanks!

PR #21, opened 2017-12-06
[codestyle] works with all versions of java

akshitdewan:
replaced getClassLoader() which does not work with J2EE according to sonar. http://sonar.xwiki.org/issues/search#issues=2a3b9a97-4a6d-421f-a152-caf8f453f395

### DIFF
--- a/xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-standard/src/main/java/org/xwiki/environment/System.java
+++ b/xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-standard/src/main/java/org/xwiki/environment/System.java
@@ -130,7 +130,7 @@ public final class System
     {
         // Step 1: Initialize Component system
         EmbeddableComponentManager ecm = new EmbeddableComponentManager();
-        ecm.initialize(classLoader == null ? System.class.getClassLoader() : classLoader);
+        ecm.initialize(classLoader == null ? Thread.currentThread().getContextClassLoader() : classLoader);
 
         // Step 2: Initialize Environment
         StandardEnvironment environment;


tmortagne:
> does not work with J2EE according to sonar

What Sonar say is that you might be surprised by the way it works (but that's for people who don't know well classloader or servlets).

This is going a bit beyond codestyle but I'm OK with it (but not really for reason given by Sonar). It might improve a few use cases (even if classloader are old stuff now...) and it should not have any negative effect.

That being said I can't find the corresponding GCI task, did you claimed it ?

akshitdewan:
Sure! Did not claim a task yet, I think I was in the process of getting a task approved, but I will claim

xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-standard/src/main/java/org/xwiki/environment/System.java
```diff
@@ -130,7 +130,7 @@ public static ComponentManager initialize(File permanentDirectory, File resource
     {
         // Step 1: Initialize Component system
         EmbeddableComponentManager ecm = new EmbeddableComponentManager();
-        ecm.initialize(classLoader == null ? System.class.getClassLoader() : classLoader);
+        ecm.initialize(classLoader == null ? Thread.currentThread().getContextClassLoader() : classLoader);
```
vmassol: I don't think this is correct. This is for the standard environment, i.e. to be used outside of JavaEE.

vmassol: I've checked more and indeed if the thread context CL is not set, it'll fall back to the system CL so the code above should have the same effect as above, even though it's purely unnecessary in a non JavaEE environment IMO. Also it might be better to use the current CL instead of the system CL.

vmassol: At the very minimum this line requires a comment to explain the usage of the current thread CL.

vmassol: Some reference: https://www.javaworld.com/article/2077344/core-java/find-a-way-out-of-the-classloader-maze.html

PR #22, opened 2017-12-08
[codestyle] updated class loader

akshitdewan:


xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-standard/src/main/java/org/xwiki/environment/internal/StandardEnvironment.java
```diff
@@ -47,7 +47,7 @@
     /**
      * @see #setResourceClassLoader(ClassLoader)
      */
-    private ClassLoader resourceClassLoader = StandardEnvironment.class.getClassLoader();
+    private ClassLoader resourceClassLoader = Thread.currentThread().getContextClassLoader();
```
tmortagne: As I told you in the chat this is not the right way. You should use Thread.currentThread().getContextClassLoader() where is needed and not use it to initialize a field.

akshitdewan:
Should this issue be closed and marked as fp?

tmortagne:
I never said it was a false positive. What I'm telling you is that this is not the right fix: you should NOT initialize that field, instead you should use the context classloader is that field is not when you need it (otherwise it does not make any sense to use the **context** classloader).

akshitdewan:
Sorry @tmortagne, I misread what you said on Riot. Made another commit - is this what you meant?

tmortagne:
Exactly :)

tmortagne:

PR #23, opened 2017-12-09
[codestyle] removed deprecated listtools

akshitdewan:


vmassol:
Note that this is not about codestyle. Are you sure it's deprecated? Also I don't think we can remove this like you did. This would mean that any extension or anyone having script in wiki pages using listtool would suddenly become broken! We certainly don't want that :)

vmassol:
If we agree to remove this (and you'll need to send a mail to the devs list to discuss it) then it should be moved to a legacy module. See http://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HBackwardCompatibility

akshitdewan:
Wasn't sure what other categories are, but Thomas said codestyle is fine for gci tasks. I thought it was deprecated based on sonar and also, could not find a ListTools class on git. Sorry about removing, I figured it wasn't being used anywhere. I will take a look at the link you sent.

vmassol:
Thanks. You won't find it in xwiki git repos because it's located in the apache velocity project, as indicated by the package name:

`import org.apache.velocity.tools.generic.ListTool`

akshitdewan:
Oh haha thanks, I even  removed the import and didn't notice :)

tmortagne:
This would be a major API breakage and really out of scope for GCI.

PR #24, opened 2017-12-29
[codestyle] Call the method Thread.start() to execute the content of …

fennecus-z:
…the run() method in a dedicated thread

I fixed this issue, I'm left with two more.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/job/history/internal/ReplayJob.java
+++ b/xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/job/history/internal/ReplayJob.java
@@ -118,7 +118,7 @@ public class ReplayJob extends AbstractJob<ReplayRequest, ReplayJobStatus> imple
     {
         Job job = this.componentManager.getInstance(Job.class, record.getJobType());
         job.initialize(record.getRequest());
-        job.run();
+        job.start(request);
     }
 
     private String getTargetNamespace()


vmassol:
I'll let @tmortagne verify if the call to `run()` (execute in the current thread) instead of `start()` was intended or not.

fennecus-z:
Thanks, here is the link to the issue by the way http://sonar.xwiki.org/issues/search#issues=3ae9013c-5de2-4422-a32e-8f768f3189ba

tmortagne:
I'm not the author of this job but given the code and the fact that what is executed here is a sub-job this is a false positive IMO. Right @mflorea ?

xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/job/history/internal/ReplayJob.java
```diff
@@ -118,7 +118,7 @@ private void replay(ExtensionJobHistoryRecord record) throws ComponentLookupExce
     {
         Job job = this.componentManager.getInstance(Job.class, record.getJobType());
         job.initialize(record.getRequest());
-        job.run();
+        job.start(request);
```
mflorea: ``Job#start(Request)`` is [deprecated](https://github.com/xwiki/xwiki-commons/blob/master/xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/Job.java#L53) since a long time and if you read the deprecation message it says:

    use {@link #initialize(Request)} then {@link #run()} instead

which is exactly what this code was doing. So there's nothing to fix here. The Sonar issue is a false positive because Job extends Runnable.

vmassol: @mflorea Are you sure? What they say on Sonar is:

> The purpose of the Thread.run() and Runnable.run() methods is to execute code in a separate, dedicated thread. Calling those methods directly doesn't make sense because it causes their code to be executed in the current thread.
>
> To get the expected behavior, call the Thread.start() method instead.

So `job.run()` is the same as `Runnable.run()` and according to Sonar this doesn't start a new thread.

How does this code ensure that a new thread is started?

mflorea: Note that it doesn't make sense to replay the extension history records asynchronous unless we would replay multiple records in parallel, but that's not the case because the order is important. We replay the records one by one, to preserve the history order. Moreover, the replay job itself is run in a separate thread and it does only one thing: replays extension history records. So it wouldn't make sense to start a new thread just to wait for it, when you can simply execute the code in the current thread.

Another reason is that the replay job wants to collect the logs from all the replayed extension history records and that would be harder if the records were replayed in separate threads.

tmortagne: > How does this code ensure that a new thread is started?

As I said in my comment, we don't want to start a new thread.

tmortagne:
Closing as false positive.

xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/job/history/internal/ReplayJob.java
vmassol: @mflorea ok thanks. If the intent if not to start a new thread then it's indeed a false positive. This means we're reusing a Runnable in a slightly weird way though since Runnable are supposed to be started by a Thread normally (we're using it as an interface). Runnable javadoc says "The Runnable interface should be implemented by any class whose instances are intended to be executed by a thread.".

PR #25, opened 2018-01-06
Replaced call to old API

ApCoder123:
I just had to change the import to a different class. The method calls stay the same

As part of the Replace calls to deprecated APIs by more up to date alternatives Task


tmortagne:
Please build the module before committing anything.

tmortagne:
Again don't commit something you did not built before.

tmortagne:

PR #26, opened 2018-01-06
Used newer API

ApCoder123:
`this.features` is deprecated and `this.featuresMap` should be used instead


vmassol:
It's deprecated but it can still be used. If you remove this code then I guess it won't work anymore, will it?

vmassol:
Ideally it should be moved to the legacy module probably (http://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HBackwardCompatibility).

ApCoder123:
@vmassol I checked it out and it turns out that the method `getExtensionFeatures` used to get that features list only uses `this.featuresMap` so it doesn't really matter if we don't set the list any more, as it is never being fetched.

tmortagne:
You did not understand what this method does. It set this.features so that users of getFeatures() method get the right list.

PR #27, opened 2018-01-06
Removed Deprecated code

ApCoder123:
Updated code for Replace calls to deprecated APIs by more up to date alternatives Task


xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/Extension.java
```diff
@@ -260,7 +260,7 @@
      */
     default ExtensionId getExtensionFeature(String featureId)
     {
-        return getFeatures().contains(featureId) ? new ExtensionId(featureId, getId().getVersion()) : null;
+        return getExtensionFeatures().contains(featureId) ? new ExtensionId(featureId, getId().getVersion()) : null;
```
tmortagne: This does not work. getFeatures() and getExtensionFeatures() don't return the same thing.

PR #28, opened 2018-01-06
Added Call to Newer API

ApCoder123:
As Part of Replace calls to deprecated APIs by more up to date alternatives


xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/EscapeTool.java
```diff
@@ -28,7 +28,7 @@
 import org.apache.commons.codec.net.BCodec;
 import org.apache.commons.codec.net.QCodec;
 import org.apache.commons.codec.net.QuotedPrintableCodec;
-import org.apache.commons.lang3.StringEscapeUtils;
+import org.apache.commons.text.StringEscapeUtils
```
tmortagne: Please always build the module before committing anything, you will see that this is valid Java code.

tmortagne:
Still not enough.

tmortagne:

PR #29, opened 2018-01-09
Exposed hints array

gonzaloUr:
issues:

http://sonar.xwiki.org/issues/search#issues=3976be71-0831-4a12-95b8-1cc0481a3332
http://sonar.xwiki.org/issues/search#issues=930c8652-4275-42ff-9b6c-3c9b09c95484


tmortagne:
This is not great from performance point of view but there is not much other choice with arrays. I guess ideally the array based methods should be deprecated and List based ones should be introduced since you can protect a List instance from write access.

PR #31, opened 2018-01-09
Exposed uris array

gonzaloUr:
issues:

http://sonar.xwiki.org/issues/search#issues=4bd85ae7-3f73-48ef-aefe-65fccd9eed7a
http://sonar.xwiki.org/issues/search#issues=e816c9b4-e709-40f9-baf7-37db2159b193


xwiki-commons-core/xwiki-commons-classloader/xwiki-commons-classloader-api/src/main/java/org/xwiki/classloader/URIClassLoader.java
```diff
@@ -96,7 +96,7 @@ public URIClassLoader(URI[] uris)
     public URIClassLoader(URI[] uris, URLStreamHandlerFactory handlerFactory)
     {
         super(new URL[0]);
-        this.finder = new URIResourceFinder(uris, handlerFactory);
+        this.finder = new URIResourceFinder(uris.clone(), handlerFactory);
```
tmortagne: If you look at URIResourceFinder constructor you will see that this is useless since it does not keep the table.

tmortagne:
False positives.

PR #32, opened 2018-01-11
[codestyle] Replace the synchronized class Stack by an unsynchronized one such as Deque

fennecus-z:
Hey, here is a link to the issue and my task:

http://sonar.xwiki.org/issues/search#issues=4d594de5-af20-4804-ac85-55adc17d205c

https://codein.withgoogle.com/dashboard/task-instances/5241134932557824/?sp-page=2


xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-handlers/xwiki-commons-extension-handler-jar/src/main/java/org/xwiki/extension/jar/internal/handler/JarExtensionJobFinishingListener.java
```diff
@@ -149,7 +150,7 @@ private void popUninstallLevel()
         ExecutionContext context = this.execution.getContext();
 
         if (context != null) {
-            Stack<UninstalledExtensionCollection> extensions = getUninstalledExtensionCollectionStack(false);
+            Deque<UninstalledExtensionCollection> extensions = (Deque<UninstalledExtensionCollection>) getUninstalledExtensionCollectionStack(false);
```
tmortagne: That's not what Sonar suggest to do. The implementation is still a Stack here, you just cast it.

PR #33, opened 2018-01-11
[codestyle] Use a StringBuilder instead.

fennecus-z:
This is a fix for the issue at 
http://sonar.xwiki.org/issues/search#issues=bdadc72e-bba4-46b8-9a8f-f42c33efade5

My GCI task is
https://codein.withgoogle.com/dashboard/task-instances/5241134932557824/?sp-page=2


xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-maven/src/main/java/org/xwiki/extension/maven/internal/MavenExtensionScanner.java
```diff
@@ -318,7 +318,9 @@ public void guess(Map<String, DefaultCoreExtension> extensions, Collection<URL>
                 if (coreExtension == null) {
                     String dependencyFileName = dependency.getArtifactId() + '-' + dependency.getVersion();
                     if (dependency.getClassifier() != null) {
-                        dependencyFileName += '-' + dependency.getClassifier();
+                        StringBuilder stringBuilder = new StringBuilder(dependency.getClassifier());
```
tmortagne: This code does not get the same result that the previous one. Also StringBuilder should be used as earlier to be fully accurate since there is concatenation in previous line too.

PR #34, opened 2018-01-13
Replaced deprecated method

siddhpant:


xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-api/src/main/java/org/xwiki/component/util/ObjectUtils.java
```diff
@@ -41,30 +42,22 @@ private ObjectUtils()
      * @param object1 the first object
      * @param object2 the second object
      * @return true of the provided objects are equal
-     * @deprecated use {@link java.util.Objects#equals(Object, Object)} instead
      */
-    @Deprecated
```
tmortagne: This method is deprecated because it's not needed anymore and an alternative should be used, why did you made it non deprecated anymore ?

The goal of the task is to replace the use of a deprecated method (like this method) by what is indicated in the Javadoc. That's not at all what you did here.

siddhpant: I couldn't find any file importing this one....

siddhpant: Sorrry.

PR #38, opened 2018-05-18
XCOMMONS-1419 : Provide a method to build query strings in Velocity

acotiuga:
* build new query string, add new parameter to existing query string, convert map to query string

### DIFF
--- a/xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/URLTool.java
+++ b/xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/URLTool.java
@@ -19,12 +19,15 @@
  */
 package org.xwiki.velocity.tools;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.http.NameValuePair;
 import org.apache.http.client.utils.URLEncodedUtils;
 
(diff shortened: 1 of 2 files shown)


xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/URLTool.java
```diff
@@ -58,4 +61,55 @@
...
+    /**
+     * Add a parameter to a query string.
+     * 
+     * @param queryString the initial query string
+     * @param name the name of the parameter
+     * @param value the value of the parameter
+     * @return the new query string
+     * @since 10.4
+     */
+    public String add(String queryString, String name, String value)
```
tmortagne: Since class is about URLs in general this method name should be more explicit on the fact that it's related to query strings.

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/URLTool.java
```diff
@@ -58,4 +61,55 @@
...
+
+    /**
+     * Convert a map into a query string.
+     * 
+     * @param map the map to be converted
+     * @return the converted query string
+     * @since 10.4
+     */
+
+    public String getQueryString(Map<String, String> map)
```
tmortagne: `serializeQueryString` ?

vmassol: You also need to document if the user needs to encore or not the key and value or if it's done by the API.

vmassol:
Actually we already have parseQuery() which returns a map so I guess if we want to add some new params we would need to parse, add to the map and then serialize again. That works.

vmassol:
anyway I'd still reuse the apache commons NameValuePair and URLEcnodeUtils classes for the implementation.

mflorea:
This is already supported with ``$escapetool.url({...})`` since XWiki 5.2

PR #41, opened 2018-06-16
Update from Weblate.

xwikiorgci:


vmassol:
This looks wrong

vmassol:
Actually I thought the translations were in English but they're just missing translations.

So "id" is Indonesian. Should we commit this even though there's no translation?

Also why is there commit https://github.com/xwiki/xwiki-commons/pull/41/commits/c09ef0bedeab1d6691c0d2cc7d6650dfe119ca52 which has no changes?

tmortagne:
I guess all that is the result of someone clicking on "Start new translation" but not actually translating any key.

PR #44, opened 2018-07-27
XCOMMONS-1447: XAR plugin should replace the dates with a common number

Enygma2002:
Removing date fields with `xar:format` and failing the build with `xar:verify` if date fields are present in the XML pages.

More details on the previous discussion: https://markmail.org/thread/ymwsebvr3k7voy3p


tmortagne:
Don't forget attachment date too.

Enygma2002:
Good point, @tmortagne. Added in the new commit.

PR #49, opened 2018-10-17
XCOMMONS-1509: CancelableEvents are not propagated when they are canceled

surli:
## Issue

https://jira.xwiki.org/browse/XCOMMONS-1509

## Solution

Override `matches(Event)` in AbstractCancelableEvent.

## Tests

Add a new unit test to check the new behaviour.


surli:
> In my mind in was DefaultObservationManager job to check if an event is cancelled before continuing to call listeners.

I understood, but I thought this implementation would allow more flexibility for this specific behaviour. i.e. if an user would want to trigger the listener no matter the state, then he can. 

If we put it in the DefaultObservationManager, then we consider that this behaviour is fixed and cannot be changed. 

Here it's really a choice to do IMO.

PR #51, opened 2018-11-07
XWIKI-15824: Add support for deprecated parameters in macro descriptors

atallahade:
### Issue

https://jira.xwiki.org/browse/XWIKI-15824

### Change

* Add a deprecated field in the PropertyDescriptor.


xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultPropertyDescriptor.java
```diff
@@ -240,4 +245,19 @@ public Method getWriteMethod()
     {
         return this.writeMethod;
     }
+
+    @Override
+    public boolean isDeprecated()
+    {
+        return deprecated;
```
vmassol: missing `this`

tmortagne:
The pull request is OK but committing on commons with a platform jira issue is not great. Would be better to have a XCOMMONS issue for this pull request and a XRENDERING one for the macro bridge.

atallahade:
Thanks @tmortagne, I'll change the issues.

PR #52, opened 2018-11-08
[Misc] Allow to specify a timeoutFactor for pitest in the command line

surli:


pom.xml
```diff
@@ -124,6 +124,9 @@
          commits. Thus the idea is that the CI will set this as a system property to run it using a scheduler. It can
          also be execute locally by running "mvn install -Pquality -Dxwiki.pitest.skip=false" -->
     <xwiki.pitest.skip>true</xwiki.pitest.skip>
+    <!-- Default Pitest timeout factor. Increasing the value will increase the duration of a build
+         but might improve the mutation score. -->
+    <xwiki.pitest.timeoutFactor>1.25</xwiki.pitest.timeoutFactor>
```
vmassol: the comment is a bit misleading :)

If I reformulate it, it can be understood as the higher the timeout, the higher the test quality are.

I think it's missing the info that when there are timeouts, mutation score can be reduced. Would be interesting to link to the GH issue on descartes too I think.

PR #53, opened 2018-11-14
XCOMMONS-1527: Create an HTML Displayer interface

atallahade:


xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayer.java
```diff
@@ -0,0 +1,45 @@
...
+@Unstable
+public interface HTMLDisplayer<T>
+{
+    /**
+     * @return the html element of the class
+     */
+    String display();
+
+    /**
+     * @param type the display type
```
vmassol: what is a type?

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayer.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Display an html element based on the given class.
+ *
+ * @version $Id$
+ * @since 10.10RC1
+ */
+@Role
+@Unstable
+public interface HTMLDisplayer<T>
```
vmassol: Missing documentation for `<T>`. What is it?

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayer.java
```diff
@@ -0,0 +1,45 @@
...
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.displayer;
+
+import org.xwiki.component.annotation.Role;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Display an html element based on the given class.
```
vmassol: What is a class?

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayer.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package org.xwiki.displayer;
+
+import java.util.Map;
+
+import org.xwiki.component.annotation.Role;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Display an html element based on the class of {@link T}.
```
tmortagne: "class" is a bit misleading since `Class` is only a subset of the supported `Type`s.

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayer.java
```diff
@@ -0,0 +1,54 @@
...
+ * @since 10.10RC1
+ */
+@Role
+@Unstable
+public interface HTMLDisplayer<T>
+{
+    /**
+     * @return the html element related to the class
+     */
+    String display();
```
tmortagne: I feel that all the display method should always take a value. Displaying a form without any default value is not really a use case we have currently (properties and macro parameters have a default value) and if you really want that then you can pass null I think. It feel much more natural that constantly have to create a Map and pass the default value as it it was some custom configuration.

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayer.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * @param parameters parameters used while generating the html. Could be the attributes of an input for instance.
+     * @return the html element related to the class
+     */
+    String display(Map<String, String> parameters);
+
+    /**
+     * @param parameters parameters used while generating the html. Could be the attributes of an input for instance.
+     * @param type the display type (view, edit, ...)
```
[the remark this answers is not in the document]
atallahade: I also thought about "action" but "mode" is actually more explicit, thanks.

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayer.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+@Role
+@Unstable
+public interface HTMLDisplayer<T>
+{
+    /**
+     * @param parameters parameters used while generating the html. Could be the attributes of an input for instance.
+     * @return the html element related to the class
+     */
+    String display(Map<String, String> parameters);
```
tmortagne: I think you misunderstood my previous comment. It's fine to have a display method without the custom parameters but as I said it should tale a value (as in `T value`) IMO since that's what the displayer is mainly about.

atallahade: @tmortagne I'm not sure to understand. You want something like this?

```suggestion
    <E> String display(E value);
```

atallahade: Ok so you want me to do:

```suggestion
    String display(T value, Map<String, String> parameters);
```

tmortagne: For this method yes.

atallahade: @tmortagne I've made the change

atallahade:
@tmortagne I've added an HTMLDisplayer Manager as you've suggested.

atallahade:
See https://github.com/xwiki/xwiki-platform/pull/952 for an implementation.

xwiki-commons-core/xwiki-commons-displayer/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.xwiki.commons</groupId>
+    <artifactId>xwiki-commons-core</artifactId>
+    <version>10.10-SNAPSHOT</version>
```
tmortagne: Don't forget to update the version.

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayer.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.xwiki.component.annotation.Role;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Display an html element based on the type of {@link T}.
+ *
+ * @param <T> type used for the display
+ * @version $Id$
+ * @since 10.10RC1
```
tmortagne: Don't forget to update the version.

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayerException.java
```diff
@@ -0,0 +1,41 @@
...
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.displayer;
+
+/**
+ * Base exception for the HTML Displayer.
+ *
+ * @version $Id$
+ * @since 10.10RC1
```
tmortagne: Don't forget to update the version.

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayerManager.java
```diff
@@ -0,0 +1,46 @@
...
+import java.lang.reflect.Type;
+
+import org.xwiki.component.annotation.Role;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Provides a way to access the HTML Displayer for the given type.
+ *
+ * @version $Id$
+ * @since 10.10RC1
```
tmortagne: Don't forget to update the version.

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/internal/DefaultDisplayerManager.java
```diff
@@ -0,0 +1,69 @@
...
+import org.xwiki.component.util.DefaultParameterizedType;
+import org.xwiki.displayer.HTMLDisplayer;
+import org.xwiki.displayer.HTMLDisplayerException;
+import org.xwiki.displayer.HTMLDisplayerManager;
+
+/**
+ * Default implementation for {@link org.xwiki.displayer.HTMLDisplayerManager}.
+ *
+ * @version $Id$
+ * @since 10.10RC1
```
tmortagne: Don't forget to update the version.

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayerManager.java
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Retrieve the appropriate HTML Displayer based on the given type.
+     *
+     * @param <T> the type of the HTML Displayer
+     * @param targetType the type used to retrieve the HTML Displayer
+     * @return the HTML Displayer of type {@code <T>}
+     * @throws HTMLDisplayerException if the HTML Displayer cannot be found
+     */
+    <T> HTMLDisplayer<T> getHTMLDisplayer(Type targetType) throws HTMLDisplayerException;
+}
```
tmortagne: I was thinking that it would be nice if that entry point reproduced the methods of HTMLDisplayer but with an additional Type parameter since that's what people will do most of the time.

atallahade:
@tmortagne is this PR ok?

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/internal/DefaultDisplayerManager.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public <T> HTMLDisplayer<T> getHTMLDisplayer(Type targetType) throws HTMLDisplayerException
+    {
+        try {
+            ComponentManager componentManager = this.componentManagerProvider.get();
+
+            ParameterizedType converterType = new DefaultParameterizedType(null, HTMLDisplayer.class, targetType);
+
+            return componentManager.getInstance(converterType);
```
tmortagne: What could be interesting is to try with the Type and then if the Type is a ParameterizedType try the Class. For example in case of List we will probably have the same implementation for most of the types so it's nice to fallback on `HTMLDisplayer<List>` if no `HTMLDisplayer<List<String>>` is not registered.

Also I feel we should fallback on `HTMLDisplayer` (with no associated type) in the end which will probably be implemented as a simple input string in edit and deal with String<->T using ConverterManager mode. WDYT ?

In short I think we should throw an exception only if the asked displayer actually exist but could not get initialized for some reason.

You can take a look at https://github.com/xwiki/xwiki-commons/blob/master/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultConverterManager.java#L128 for inspiration. On important thing to notice if the use of hasComponent which is better for performance when there is a good chance the component does not exist and allow making the difference between the document does not exist and there is a problem with the document.

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayer.java
```diff
@@ -0,0 +1,61 @@
...
+ * @param <T> type used for the display
+ * @version $Id$
+ * @since 10.11RC1
+ */
+@Role
+@Unstable
+public interface HTMLDisplayer<T>
+{
+    /**
+     * @param value the value on which the display is based on
```
mflorea:
```suggestion
     * @param value the value to be displayed
```

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayer.java
```diff
@@ -0,0 +1,61 @@
...
+     * @return the html element based on the value and the type
```
mflorea:
```suggestion
     * @return the HTML that can be used to display the given value
```
First of all, the returned string could have multiple HTML "elements" and it could contain comments or other HTML syntax. We could have for instance:

    <input type="checkbox" name="enabled" value="true" />
    
    <input type="hidden" name="enabled" value="false" />

Then, from what I know, "HTML element" is used mostly when your talk about **object** representation (e.g. DOM element), while "HTML tag" is used to denote the **string** representation. So you can say "return the HTML tags" or "return the HTML markup" or simply "return the HTML".

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayerManager.java
```diff
@@ -0,0 +1,89 @@
...
+package org.xwiki.displayer;
+
+import java.lang.reflect.Type;
+import java.util.Map;
+
+import org.xwiki.component.annotation.Role;
+import org.xwiki.stability.Unstable;
+
+/**
+ * Provides a way to access the HTML Displayer for the given type.
```
mflorea: You can access the HTML displayer for a given type by injection or using the component manager. Why should I use this manager instead?

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/internal/DefaultDisplayerManager.java
```diff
@@ -0,0 +1,123 @@
...
+     */
+    @Override
+    public <T> HTMLDisplayer<T> getHTMLDisplayer(Type targetType, String roleHint) throws HTMLDisplayerException
+    {
+        try {
+            HTMLDisplayer<T> component;
+            ComponentManager componentManager = this.componentManagerProvider.get();
+
+            Type type = targetType;
+            Type converterType = new DefaultParameterizedType(null, HTMLDisplayer.class, type);
```
mflorea: Why ``converter``? Why not ``displayerType``?

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayer.java
atallahade: I agree, "return the HTML markup" or "return the HTML" is more accurate.

xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/HTMLDisplayerManager.java
atallahade: So you don't have to reimplement the logic of `DefaultDisplayerManager#getHTMLDisplayer` each time you want to retrieve an HTML displayer:
https://github.com/xwiki/xwiki-commons/blob/c0868ea92abe1b1fbd14e6e1047e83bdefd61ae9/xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/internal/DefaultDisplayerManager.java#L62-L103

tmortagne: I think what @mflorea meant is that it's not the only way to access the HTML Displayer if you already known which one you want. It's just a helper.

mflorea: Yes, my question was more for the Javadoc. I would mention that if you know precisely what type you want then you can inject directly the corresponding displayer. Otherwise, if you need the fall back on the generic displayer then you should use this component.

PR #54, opened 2018-11-15
XCOMMONS-1528: Add support for grouped parameters in property descriptors

atallahade:
### Issue

https://jira.xwiki.org/browse/XCOMMONS-1528

### Changes

* Create a PropertyGroups annotation to set a hierarchy of groups on a
  property.
* Add related tests.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyDescriptor.java
+++ b/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyDescriptor.java
@@ -22,6 +22,8 @@ package org.xwiki.properties;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
+import java.util.Collections;
+import java.util.List;
 
 /**
  * Describe a property in a bean.
@@ -108,5 +110,13 @@ public interface PropertyDescriptor
      */
     default boolean isAdvanced() {
         return false;
+    }
+
+    /**
+     * @return the hierarchy of groups.
+     * @since 10.10RC1
+     */
+    default List<String> getGroups() {
+        return Collections.emptyList();
     }
 }
--- a/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyDescriptor.java
+++ b/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyDescriptor.java
@@ -23,7 +23,6 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
 import java.util.Collections;
-import java.util.List;
 
 /**
  * Describe a property in a bean.
@@ -116,7 +115,7 @@ public interface PropertyDescriptor
      * @return the hierarchy of groups.
      * @since 10.10RC1
      */
-    default List<String> getGroups() {
+    default PropertyGroupContainer getGroup() {
         return Collections.emptyList();
     }
 }
--- a/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyDescriptor.java
+++ b/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyDescriptor.java
@@ -113,7 +113,7 @@ public interface PropertyDescriptor
 
     /**
      * @return the hierarchy of groups.
-     * @since 10.10RC1
+     * @since 10.11RC1
      */
     default PropertyGroupDescriptor getGroupDescriptor() {
         return new PropertyGroupDescriptor(Collections.emptyList());
--- a/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyDescriptor.java
+++ b/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyDescriptor.java
@@ -115,7 +115,7 @@ public interface PropertyDescriptor
      * @return the hierarchy of groups.
      * @since 10.10RC1
      */
-    default PropertyGroupContainer getGroup() {
-        return Collections.emptyList();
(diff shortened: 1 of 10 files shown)


xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/annotation/PropertyGroups.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Use this annotation to group properties.
+ *
+ * @version $Id$
+ * @since 10.10RC1
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.METHOD, ElementType.FIELD })
+@Inherited
+public @interface PropertyGroups
```
tmortagne: I think it should be PropertyGroup, what you pass is not exactly several groups but a group path as a String[] instead of a String with / that's all.

atallahade: Indeed, thanks.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultPropertyDescriptor.java
```diff
@@ -91,6 +92,11 @@
      */
     private boolean advanced;
 
+    /**
+     * @see #getGroups()
+     */
+    private List<String> groups;
```
tmortagne: Same comment than previously about `groups` vs `group`.

Instead of using directly a `List<String>` maybe it would be clearer to use some `PropertyGroup` (which contain the list) or a different name to not conflict with the annotation name.

atallahade: Ok I see.

atallahade: @tmortagne not sure about the name: `PropertyGroupContainer`.

tmortagne: Definitely not `PropertyGroupContainer` IMO since it's IS a group. If `PropertyGroup` in a different package than the annotation then I guess `PropertyGroupDescriptor` is fine, consistent with other *Descriptor classes..

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyDescriptor.java
```diff
@@ -109,4 +110,12 @@ default boolean isDeprecated() {
     default boolean isAdvanced() {
         return false;
     }
+
+    /**
+     * @return the hierarchy of groups.
+     * @since 10.10RC1
```
tmortagne: Will need to change the `@since`.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyGroupDescriptor.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+package org.xwiki.properties;
+
+import java.util.List;
+
+/**
+ * Contains a {@code List<String>} to view the group as a single object.
+ *
+ * @version $Id$
+ * @since 10.10RC1
```
tmortagne: Will need to change the `@since`.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/annotation/PropertyGroup.java
```diff
@@ -0,0 +1,43 @@
...
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Use this annotation to group properties.
+ *
+ * @version $Id$
+ * @since 10.10RC1
```
tmortagne: Will need to change the `@since`.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultPropertyDescriptor.java
```diff
@@ -280,4 +286,19 @@ public void setAdvanced(boolean advanced)
...
+    public PropertyGroupDescriptor getGroupDescriptor()
+    {
+        return this.group;
+    }
+
+    /**
+     * @param group the property hierarchy of groups
+     * @see #getGroupDescriptor()
+     */
+    public void setGroup(PropertyGroupDescriptor group)
```
tmortagne: I would add a `@since`.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyGroupDescriptor.java
```diff
@@ -0,0 +1,51 @@
...
+    private List<String> group;
+
+    /**
+     * Default constructor.
+     *
+     * @param group hierarchy of groups
+     */
+    public PropertyGroupDescriptor(List<String> group)
+    {
+        this.group = group;
```
tmortagne: Would be safer to do something like

`this.groups = Collections.unmodifiableList(group)`

otherwise anyone can modify the actual path trough getGroup() which is not great for a public API.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyDescriptor.java
atallahade: Sure.

PR #55, opened 2018-11-23
XCOMMONS-1530: Add support for feature parameters in property descriptors

atallahade:
### Issue

https://jira.xwiki.org/browse/XCOMMONS-1530

### Changes

* Create a PropertyFeature annotation.
* Add an feature field in the PropertyDescriptor.
* Add related tests.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyDescriptor.java
+++ b/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyDescriptor.java
@@ -118,4 +118,12 @@ public interface PropertyDescriptor
     default PropertyGroupDescriptor getGroupDescriptor() {
         return new PropertyGroupDescriptor(Collections.emptyList());
     }
+
+    /**
+     * @return the name of the feature associated to the parameter.
+     * @since 10.11RC1
+     */
+    default String getFeature() {
+        return null;
+    }
 }
--- a/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyDescriptor.java
+++ b/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyDescriptor.java
@@ -22,7 +22,6 @@ package org.xwiki.properties;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
-import java.util.Collections;
 
 /**
  * Describe a property in a bean.
@@ -116,14 +115,6 @@ public interface PropertyDescriptor
      * @since 10.11RC1
      */
     default PropertyGroupDescriptor getGroupDescriptor() {
-        return new PropertyGroupDescriptor(Collections.emptyList());
-    }
-
-    /**
-     * @return the name of the feature associated to the parameter.
-     * @since 10.11RC1
-     */
-    default String getFeature() {
-        return null;
+        return new PropertyGroupDescriptor(null);
     }
 }
--- a/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyGroupDescriptor.java
+++ b/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyGroupDescriptor.java
@@ -23,7 +23,7 @@ import java.util.Collections;
 import java.util.List;
 
 /**
- * Contains a {@code List<String>} to view the group as a single object.
+ * Contains a {@code List<String>} to view the group as a single object and the associated feature.
  *
  * @version $Id$
  * @since 10.11RC1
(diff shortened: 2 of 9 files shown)


xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyDescriptor.java
```diff
@@ -118,4 +118,12 @@ default boolean isAdvanced() {
     default PropertyGroupDescriptor getGroupDescriptor() {
         return new PropertyGroupDescriptor(Collections.emptyList());
     }
+
+    /**
+     * @return the name of the feature associated to the parameter.
+     * @since 10.11RC1
+     */
+    default String getFeature() {
```
tmortagne: For me the whole point was to make this a property group thing so located in the PropertyGroupDescriptor.

atallahade: Sure, good idea.

tmortagne: I think it also mean that the properties you share the same instance of PropertyGroupDescriptor, otherwise you could end up with properties with the same group but different feature.

atallahade: @tmortagne  What happen if a field is annotate with `@PropertyFeature` and not `@PropertyGroup`? Should the group in the `PropertyGroupDescriptor` be null, or maybe an empty list? WDYT?

tmortagne: I would go for null I think, makes more sense to me.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyGroupDescriptor.java
```diff
@@ -32,21 +32,43 @@
 {
     private List<String> group;
 
+    private String feature;
```
tmortagne: You don't like the multi features idea ? :)

May be hard to add later without breaking the API if we have the need.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyGroupDescriptor.java
```diff
@@ -32,21 +32,43 @@
...
+
     /**
      * Default constructor.
      *
-     * @param group hierarchy of groups
+     * @param group hierarchy of groups or null
```
tmortagne: You should probably give a little more details on what null mean exactly.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultBeanDescriptor.java
```diff
@@ -66,7 +68,9 @@
     /**
      * The properties of the bean.
      */
-    private Map<String, PropertyDescriptor> parameterDescriptorMap = new LinkedHashMap<String, PropertyDescriptor>();
+    private Map<String, PropertyDescriptor> parameterDescriptorMap = new LinkedHashMap<>();
+
+    private Map<PropertyGroup, PropertyGroupDescriptor> groups = new HashMap<>();
```
tmortagne: You sure two different instances of PropertyGroup with the same value give the same hashCode/equals ?

atallahade: [This](https://docs.oracle.com/javase/8/docs/api/java/lang/annotation/Annotation.html#equals-java.lang.Object-) + the tests would have failed otherwise.

tmortagne: OK it's surprising but that's good to know :)

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/PropertyGroupDescriptor.java
atallahade: There is actually a check done when 2 groups have a different feature assigned:
https://github.com/xwiki/xwiki-commons/blob/85199240eb14f223642c5df00e308329ee2ec35f/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultBeanDescriptor.java#L279-L284

It's just a warning here but we can even throw an exception if we want to enforce the rule (at most one feature per group).

As for multi-features, I don't quite see how it fits our need and I think it will actually bring more complexity.

tmortagne: An error would be safer otherwise the `feature` which end up chosen is going to be random (since the order of the fields/methods is not guaranteed).

tmortagne: I mean an exception.

PR #56, opened 2018-11-28
XCOMMONS-1527: Create an HTML Displayer interface

atallahade:
### Issue

https://jira.xwiki.org/browse/XCOMMONS-1527

### Changes

* Improve the implementation of the default HTMLDisplayerManager.
* Adapt and refactor the tests.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/internal/DefaultHTMLDisplayerManager.java
+++ b/xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/internal/DefaultHTMLDisplayerManager.java
@@ -44,7 +44,7 @@ import org.xwiki.displayer.HTMLDisplayerManager;
  */
 @Component
 @Singleton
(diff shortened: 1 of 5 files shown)


xwiki-commons-core/xwiki-commons-displayer/src/main/java/org/xwiki/displayer/internal/DefaultHTMLDisplayerManager.java
```diff
@@ -105,19 +105,31 @@
     @Override
     public <T> String display(Type targetType, T value) throws HTMLDisplayerException
     {
-        return getHTMLDisplayer(targetType).display(value);
+        if (targetType == null && value != null) {
```
tmortagne: Just a detail: I would have probably created a private (maybe it make sense even in public not sure) getHTMLDisplayer(Type targetType, Object value) to put this logic in a common method.

atallahade: Right, I'll refactor

PR #57, opened 2018-12-18
XWIKI-15978: Make shutdown much faster by optimizing sorting of component dependencies in dispose()

m-lw:
Here is a pull request for the patch in https://jira.xwiki.org/browse/XWIKI-15978.


atallahade:
Just tested locally and it works fine. My wiki takes 2sec to shutdown versus ~20sec before the change :+1:. Merging it as discussed with @tmortagne.

Thanks for your contribution @m-lw!!

PR #59, opened 2018-12-19
[Misc] Add a getter for AbstractRequest#properties

surli:


xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/AbstractRequest.java
```diff
@@ -246,4 +247,13 @@ public void setContext(Map<String, Serializable> context)
...
         setProperty(PROPERTY_CONTEXT, context);
     }
+
+    /**
+     * @return the map of properties in an {@link UnmodifiableMap}.
+     * @since 10.11
+     */
+    public Map<String, Object> getProperties()
+    {
+        return UnmodifiableMap.decorate(properties);
```
tmortagne: Not sure what is the benefit compared to `Collections.unmodifiableMap`.

surli: yeah sorry I'm a bit like jenkins today, in a bad shape. :sneezing_face:

PR #60, opened 2019-01-04
WIP XCOMMONS-1551: HtmlCleaner shouldn't transformed content when creating a Document

surli:
* Add new tests

### DIFF
--- a/xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
@@ -28,6 +28,7 @@ import java.util.Map;
 
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
+import org.w3c.dom.Document;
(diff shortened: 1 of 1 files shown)


surli:
Do you agree with this test @tmortagne ? It's not passing yet, but I want to make sure that we are on the same page for its contract.

tmortagne:
> Do you agree with this test @tmortagne ? It's not passing yet, but I want to make sure that we are on the same page for its contract.

The following check

> assertEquals("http://host.com/a.gif?a=foo&amp;b=bar", textContent);

suggest that you still expect to find an HTML entity in the Document content. Did I missed something ?

I'm referring to https://github.com/xwiki/xwiki-commons/pull/60/commits/1ea223d8f4729070d0bc0fb7beee1db602da586e#diff-662b1d8baf4677911842867688c409bfR120, that's the test you are talking about, right ?

surli:
>>    assertEquals("http://host.com/a.gif?a=foo&amp;b=bar", textContent);

> suggest that you still expect to find an HTML entity in the Document content. Did I missed something ?

The HTML input is: 
```
<img src="http://host.com/a.gif?a=foo&amp;b=bar" />
```
so it checks that indeed this input is not transformed.
But as I said in my last comments, here those entities are always deserialized by HtmlCleaner.

surli:
Closing this in favor of #93

PR #61, opened 2019-01-07
XCOMMONS-1553: Remove XWikiDOMSerializer in favor of HtmlCleaner DomSerializer

surli:
## Issue

https://jira.xwiki.org/browse/XCOMMONS-1553

## Solution
  * Remove XWikiDOMSerializer, use HtmlCleaner DomSerializer and fix
tests accordingly.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-xml/pom.xml
+++ b/xwiki-commons-core/xwiki-commons-xml/pom.xml
@@ -32,7 +32,7 @@
   <packaging>jar</packaging>
   <description>XWiki Commons - XML</description>
   <properties>
-    <xwiki.jacoco.instructionRatio>0.61</xwiki.jacoco.instructionRatio>
+    <xwiki.jacoco.instructionRatio>0.59</xwiki.jacoco.instructionRatio>
     <xwiki.pitest.mutationThreshold>63</xwiki.pitest.mutationThreshold>
     <!-- There's a utility class with lots of features, allow it to have many dependencies;
          There's a SAX event listener, which requires complex code -->
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/checkstyle/checkstyle-suppressions.xml
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/checkstyle/checkstyle-suppressions.xml
@@ -32,4 +32,8 @@
   <suppress checks="ClassFanOutComplexity" files="SAXEventWriter.java" />
   <suppress checks="CyclomaticComplexity" files="SAXEventWriter.java" />
   <suppress checks="AnonInnerLength" files="SAXEventWriter.java" />
+
+  <!-- XWikiDomSerializer copied from DomSerializer -->
+  <suppress checks="CyclomaticComplexity" files="XWikiDOMSerializer" />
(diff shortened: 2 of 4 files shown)


surli:
:warning: Note that removing the XWikiDOMSerializer led to a decrease in the coverage which makes Jacoco fails the build.
I'll improve coverage once some reviewers agree on those changes

surli:
@vmassol @tmortagne could you review this one?
There's one thing I'm not sure about: it's the encoding of entities when CDATA is commented. Before even if CDATA was commented, the entities were not encoded, right now they are. See: https://github.com/xwiki/xwiki-commons/pull/61/files#diff-662b1d8baf4677911842867688c409bfR206

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
```diff
@@ -155,7 +155,7 @@ public Document clean(Reader originalHtmlContent, HTMLCleanerConfiguration confi
             cleanedNode.setDocType(new DoctypeToken("html", "PUBLIC", "-//W3C//DTD XHTML 1.0 Strict//EN",
                 "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd"));
             result =
-                new XWikiDOMSerializer(cleanerProperties).createDOM(getAvailableDocumentBuilder(), cleanedNode);
+                new DomSerializer(cleanerProperties).createDOM(cleanedNode);
```
tmortagne: Note that not reusing the context DocumentBuilder will have a performance impact but I guess there is not much we can do about it.

surli: >  I guess there is not much we can do about it.

I just checked and indeed the document builder is created inside the method so I don't really see how to provide it.

surli:
@vmassol any feedback on this one?

vmassol:
> @vmassol @tmortagne could you review this one?
> There's one thing I'm not sure about: it's the encoding of entities when CDATA is commented. Before even if CDATA was commented, the entities were not encoded, right now they are. See: https://github.com/xwiki/xwiki-commons/pull/61/files#diff-662b1d8baf4677911842867688c409bfR206

Sounds good and indeed sounds like it would even correct a current bug we have.

vmassol:
Only (important) issue is the loss of performance.

atallahade:
@vmassol 
> Only (important) issue is the loss of performance.

Maybe we should do some performance tests to make sure it's not too bad, WDYT?

tmortagne:
> Maybe we should do some performance tests to make sure it's not too bad, WDYT?

Note that creating the DocumentBuilder is not super fast but it's specifically more about the DocumentBuilderFactory creation mainly because it's blocking all the threads which are doing stuff implying loading a class from the classloader making it an important lock contention. I modified its behavior (and other similar tasks) after noticing that there was often a bunch of threads waiting for this kind of lock.

PR #63, opened 2019-01-22
XWIKI-16036: Random warning log - Could not find any matching step

surli:
* Add a check to avoid a NPE in
DefaultObservationManager#removeListener

### DIFF
--- a/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-local/src/main/java/org/xwiki/observation/internal/DefaultObservationManager.java
+++ b/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-local/src/main/java/org/xwiki/observation/internal/DefaultObservationManager.java
@@ -221,6 +221,10 @@ public class DefaultObservationManager implements ObservationManager
     @Override
     public void removeListener(String listenerName)
     {
+        // this might happen in some tests
+        if (this.listenersByEvent == null) {
+            initializeListeners();
+        }
         synchronized (this.listenersByEvent) {
             getListenersByName().remove(listenerName);
             for (Map.Entry<Class<? extends Event>, Map<String, RegisteredListener>> entry : this.listenersByEvent
--- a/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-local/src/main/java/org/xwiki/observation/internal/DefaultObservationManager.java
+++ b/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-local/src/main/java/org/xwiki/observation/internal/DefaultObservationManager.java
@@ -221,17 +221,15 @@ public class DefaultObservationManager implements ObservationManager
     @Override
     public void removeListener(String listenerName)
     {
-        // this might happen in some tests
-        if (this.listenersByEvent == null) {
-            initializeListeners();
-        }
-        synchronized (this.listenersByEvent) {
-            getListenersByName().remove(listenerName);
-            for (Map.Entry<Class<? extends Event>, Map<String, RegisteredListener>> entry : this.listenersByEvent
-                .entrySet()) {
-                entry.getValue().remove(listenerName);
-                if (entry.getValue().isEmpty()) {
-                    this.listenersByEvent.remove(entry.getKey());
+        if (this.listenersByEvent != null) {
+            synchronized (this.listenersByEvent) {
+                getListenersByName().remove(listenerName);
+                for (Map.Entry<Class<? extends Event>, Map<String, RegisteredListener>> entry : this.listenersByEvent
+                    .entrySet()) {
+                    entry.getValue().remove(listenerName);
+                    if (entry.getValue().isEmpty()) {
+                        this.listenersByEvent.remove(entry.getKey());
+                    }
                 }
             }
         }
--- a/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-local/src/main/java/org/xwiki/observation/internal/DefaultObservationManager.java
+++ b/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-local/src/main/java/org/xwiki/observation/internal/DefaultObservationManager.java
@@ -131,7 +131,9 @@ public class DefaultObservationManager implements ObservationManager
             initializeListeners();
         }
 
-        return this.listenersByEvent;
+        synchronized (this.listenersByEvent) {
+            return this.listenersByEvent;
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-local/src/main/java/org/xwiki/observation/internal/DefaultObservationManager.java
```diff
@@ -221,6 +221,10 @@ public void addListener(EventListener eventListener)
     @Override
     public void removeListener(String listenerName)
     {
+        // this might happen in some tests
+        if (this.listenersByEvent == null) {
+            initializeListeners();
+        }
```
vmassol: hmm this doesn't look very nice.... First the comment is not very appealing and then initializing listeners to remove one feels weird :)

vmassol: PS: Small detail: Would be great if you could start your sentences with an uppercase in comments :)

vmassol: It's also not good to change runtime code for tests. The justification is not good enough IMO.

surli: I'll encapsulate the entire method in a `if (this.listenersByEvent != null)` to avoid the NPE. But I'll also propose a small change in the test since basically it proves that it calls the remove even if there's no listeners yet.

vmassol: > it proves that it calls the remove even if there's no listeners yet.

That's not a bad test actually. It's the implementation that seems wrong. Why remove a listener if there are none? I think that's what you meant by surrounding the whole method with an `if`.

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-local/src/main/java/org/xwiki/observation/internal/DefaultObservationManager.java
```diff
@@ -221,13 +221,15 @@ public void addListener(EventListener eventListener)
...
     public void removeListener(String listenerName)
     {
-        synchronized (this.listenersByEvent) {
-            getListenersByName().remove(listenerName);
-            for (Map.Entry<Class<? extends Event>, Map<String, RegisteredListener>> entry : this.listenersByEvent
-                .entrySet()) {
-                entry.getValue().remove(listenerName);
-                if (entry.getValue().isEmpty()) {
-                    this.listenersByEvent.remove(entry.getKey());
+        if (this.listenersByEvent != null) {
```
vmassol: The other option is to not have to check for this and initialize the listenersByEvent as some empty map. I don't recall why we do this. At first glance, since there's only a single instance of DefaultObservationManager it doesn't seem like it would be a costly operation to do a `new ConcurrentHashMap<>();`.

tmortagne: This if is wrong. It makes impossible to remove one of the registered listeners until some other manipulation of the ObservationManager is done (send an event, add another listener).

You just need to lock the return of getListenersByName() instead of this.listenersByEvent directly.

Enygma2002: Why not simply initialize the 2 maps in the default constructor or in the field definitions? Why do we need to rely on the lazy `initializeListeners` method instead of making sure we always have non-null maps already created?

Even better, instead of adding an "isInitialized" flag (to replace the null check that we are using now and is having side-effects), we could simply make the component `implement Initializable` and do both the map creations and the lookup of the listeners (i.e. calling `initializeListeners`) inside the `initialize` method.

WDYT?

vmassol: Re lazy init, see https://github.com/xwiki/xwiki-commons/blob/ac8c331e133a6480bbf08c92603a6c91bbd5eb8a/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-local/src/main/java/org/xwiki/observation/internal/DefaultObservationManager.java#L150

Enygma2002: Then maybe we need to use the `isInitialized` flag, as I mentioned, instead of the null check on the maps. This frees the maps to always be non-null and we can avoid the NPE as well.

surli: actually right now I'm trying to remove the lock on the map, and to lock the getter methods on the map instead. Taking some time because I want to test it properly :)

surli:
@tmortagne so it's finally ready and tested. It still fix XWIKI-16036 and it avoids the NPE.

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-local/src/main/java/org/xwiki/observation/internal/DefaultObservationManager.java
```diff
@@ -131,7 +131,9 @@ void removeEvent(Event event)
             initializeListeners();
         }
 
-        return this.listenersByEvent;
+        synchronized (this.listenersByEvent) {
```
tmortagne: This is really not great for performance since it adds locking on read (notify). If you really need something like this then you should use a ReadWriteLock instead.

Also this lock does not have any impact on removeListener which uses getListenersByName(), are you sure it's enough ?

surli: > This is really not great for performance since it adds locking on read (notify)

actually that's why I preferred the solution to put the synchronize in add/remove listener to avoid a bit perf issues.

> Also this lock does not have any impact on removeListener which uses getListenersByName(), are you sure it's enough ?

I'm sure I did not reproduce the bug with it. But AFAIU the bug is linked with adding a listener, not removing, so it makes sense.

tmortagne: > I'm sure I did not reproduce the bug with it. But AFAIU the bug is linked with adding a listener, not removing, so it makes sense.

In previous version you did not had any locking on notify side. That would mean the issue is concurrency issue between several threads doing addListener only ?

surli: > That would mean the issue is concurrency issue between several threads doing addListener only ?

I think so, yes. That's why it appears very rarely IMO

tmortagne: OK but it also this the synchronoize at this level is not needed at all and having one at addListener level should be enough.

surli: not sure to understand: you mean I should have only one synchronize in addListener, and nothing elsewhere, right?

PR #64, opened 2019-01-23
XCOMMONS-1558: Create a PropertyDisplayType annotation

atallahade:
### Issue

https://jira.xwiki.org/browse/XCOMMONS-1558

### Changes

* Introduce a PropertyDisplayType annotation that we can use to specify with which type the macro parameter should be rendered.
* Add the display type in the property descriptor.

### Note

The idea is to put the `PropertyDisplayType` annotation on macro parameters and, in the ckeditor application, use the provided type for the input rendering instead of the actual parameter type.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/annotation/PropertyDisplayType.java
+++ b/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/annotation/PropertyDisplayType.java
@@ -29,7 +29,7 @@ import java.lang.annotation.Target;
  * Use this annotation to specify the class type of a property.
  *
  * @version $Id$
- * @since 11.0RC1
+ * @since 11.0
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ ElementType.METHOD, ElementType.FIELD })
@@ -37,7 +37,7 @@ import java.lang.annotation.Target;
 public @interface PropertyDisplayType
 {
     /**
-     * @return the class type.
+     * @return the class type (hierarchy).
      */
-    Class value();
+    Class[] value();
 }
--- a/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/annotation/PropertyDisplayType.java
+++ b/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/annotation/PropertyDisplayType.java
@@ -27,6 +27,8 @@ import java.lang.annotation.Target;
 
 /**
  * Use this annotation to specify the class type of a property.
+ * <p> You can for instance specify the generic type Map&lt;String, Long&gt; using the following value:
+ * {Map.class, String.class, Long.class}
  *
  * @version $Id$
  * @since 11.0
@@ -37,7 +39,7 @@ import java.lang.annotation.Target;
 public @interface PropertyDisplayType
 {
     /**
-     * @return the class type (hierarchy).
+     * @return the class type followed by its parameter types.
      */
     Class[] value();
 }
--- a/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/annotation/PropertyDisplayType.java
+++ b/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/annotation/PropertyDisplayType.java
@@ -25,6 +25,8 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+import org.xwiki.stability.Unstable;
+
 /**
  * Use this annotation to specify the class type of a property.
  * <p> You can for instance specify the generic type Map&lt;String, Long&gt; using the following value:
@@ -36,6 +38,7 @@ import java.lang.annotation.Target;
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ ElementType.METHOD, ElementType.FIELD })
 @Inherited
+@Unstable
 public @interface PropertyDisplayType
 {
     /**
--- a/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/annotation/PropertyDisplayType.java
+++ b/xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/annotation/PropertyDisplayType.java
@@ -1,0 +1,43 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.properties.annotation;
+
+import java.lang.annotation.ElementType;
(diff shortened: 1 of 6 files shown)


xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultBeanDescriptor.java
```diff
@@ -222,6 +234,16 @@ protected void extractPropertyDescriptor(Field field, Object defaultInstance)
...
 
+            // set parameter display type
+            PropertyDisplayType displayTypeAnnotation = field.getAnnotation(PropertyDisplayType.class);
+            Type displayType;
+            if (displayTypeAnnotation != null) {
+                displayType = displayTypeAnnotation.value();
+            } else {
+                displayType = desc.getPropertyType();
+            }
+            desc.setDisplayType(displayType);
```
mflorea: There's a bit of code duplication here, not much, but I would still put it in a separate private method.

tmortagne:
I don't see the actual PropertyDisplayType annotation definition. Is it in a different PR ? Missing git add ?

atallahade:
@tmortagne Indeed the file is missing. I will also fix some checkstyle issues and write some tests.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/annotation/PropertyDisplayType.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Use this annotation to specify the class type of a property.
+ *
+ * @version $Id$
+ * @since 11.0RC1
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.METHOD, ElementType.FIELD })
+@Inherited
+public @interface PropertyDisplayType
```
tmortagne: This is not enough IMO since you only cover Class use case. Problem is that Type is not allowed in an annotation but at the very least you should add optional Class[] parameters field to cover things like List<Integer> (and update descriptor code accordingly), that should be enough for 99% use cases.

atallahade: I've made the improvement, thanks for the suggestion.

tmortagne: I feel having everything in `values` might be a bit confusing. At the very least the javadoc should be a bit more details on how to use it depending on the need.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultBeanDescriptor.java
```diff
@@ -286,6 +286,22 @@ private void setCommonProperties(DefaultPropertyDescriptor desc, Map<Class, Anno
...
+
+        PropertyDisplayType displayTypeAnnotation = (PropertyDisplayType) annotations.get(PropertyDisplayType.class);
+        Type displayType;
+        if (displayTypeAnnotation != null && displayTypeAnnotation.value().length > 0) {
+            Class[] types = displayTypeAnnotation.value().clone();
+            ArrayUtils.reverse(types);
+            displayType = types[0];
+
+            for (int i = 1; i < types.length; i++) {
+                displayType = new DefaultParameterizedType(null, types[i], displayType);
```
tmortagne: Using a `for` does not make any sense to me. With that code it's impossible to have several parameters, like Map<String, Integer> for example.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/annotation/PropertyDisplayType.java
atallahade: I agree that the javadoc should be improved, at least with examples, but I think this is actually better than adding another field. For instance you can write:
```
@PropertyDisplayType(Boolean.class)
```
=> `Boolean`
and
```
@PropertyDisplayType({List.class, Boolean.class})
```
=> `List<Boolean>`

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultBeanDescriptor.java
```diff
@@ -286,6 +286,22 @@ private void setCommonProperties(DefaultPropertyDescriptor desc, Map<Class, Anno
             }
             group.setFeature(parameterFeature.value());
         }
+
+
+        PropertyDisplayType displayTypeAnnotation = (PropertyDisplayType) annotations.get(PropertyDisplayType.class);
+        Type displayType;
+        if (displayTypeAnnotation != null && displayTypeAnnotation.value().length > 0) {
+            Class[] types = displayTypeAnnotation.value().clone();
+            ArrayUtils.reverse(types);
```
tmortagne: I don't understand what reverse is doing here. You should take the first element as raw type (List in your uni test) and then create a sub table with the rest for the parameters.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultBeanDescriptor.java
```diff
@@ -286,6 +286,22 @@ private void setCommonProperties(DefaultPropertyDescriptor desc, Map<Class, Anno
             }
             group.setFeature(parameterFeature.value());
         }
+
+
+        PropertyDisplayType displayTypeAnnotation = (PropertyDisplayType) annotations.get(PropertyDisplayType.class);
+        Type displayType;
+        if (displayTypeAnnotation != null && displayTypeAnnotation.value().length > 0) {
```
tmortagne: If there is only one Class in the table you should use it directly. Creating a DefaultParameterizedType with no arguments would not make much sense.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultBeanDescriptor.java
atallahade: Right, I was thinking of having just a hierarchy of classes. Do you think we should restrict the type to be of the form `Class1<Class2, Class3, ...>`?

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultBeanDescriptor.java
atallahade: The idea was as follow:
```
@PropertyDisplayType({Class1.class, Class2.class, Class3.class})
```
=> `Class1<Class2<Class3>>>`

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultBeanDescriptor.java
atallahade: this is the case.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultBeanDescriptor.java
tmortagne: > Do you think we should restrict the type to be of the form Class1<Class2, Class3, ...>

That's what I meant with 99%, 1% is stuff like `Map<String, List<Integer>>` I think. If there is really a string need for it we can think about it later, not easy to express in an annotation.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultBeanDescriptor.java
tmortagne: OK. For me `Class1<Class2<Class3>>>` is way more rare than `Class1<Class2, Class3>`.

xwiki-commons-core/xwiki-commons-properties/src/main/java/org/xwiki/properties/internal/DefaultBeanDescriptor.java
tmortagne: Indeed I looked too quickly.

PR #67, opened 2019-01-24
[Misc] Fix mutation threshold in commons-xml

surli:
* Add a new test class for AbstractHTMLFilter

### DIFF
--- a/xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/html/filter/AbstractHTMLFilterTest.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/html/filter/AbstractHTMLFilterTest.java
@@ -40,9 +40,9 @@ import org.xwiki.xml.internal.html.filter.LinkFilter;
 import org.xwiki.xml.internal.html.filter.ListFilter;
 import org.xwiki.xml.internal.html.filter.ListItemFilter;
 
-import static org.junit.Assert.assertTrue;
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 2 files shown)


xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/html/filter/AbstractHTMLFilterTest.java
```diff
@@ -0,0 +1,183 @@
...
+import org.xwiki.xml.html.HTMLCleaner;
+import org.xwiki.xml.internal.html.DefaultHTMLCleaner;
+import org.xwiki.xml.internal.html.filter.AttributeFilter;
+import org.xwiki.xml.internal.html.filter.BodyFilter;
+import org.xwiki.xml.internal.html.filter.FontFilter;
+import org.xwiki.xml.internal.html.filter.LinkFilter;
+import org.xwiki.xml.internal.html.filter.ListFilter;
+import org.xwiki.xml.internal.html.filter.ListItemFilter;
+
+import static org.junit.Assert.assertTrue;
```
vmassol: This is not correct. Shoud be jupiter

vmassol:
I would have written the test differently, just testing the filter and not htmlcleaner.

vmassol:
Actually I think I would have written unit tests for the extending filters instead.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/html/filter/AbstractHTMLFilterTest.java
surli: Fixed.

PR #69, opened 2019-01-28
HTMLCleaner doesn't recognize some HTML named entity

surli:
- Fix the pattern to recognize entities
  - Add one more test

### DIFF
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
@@ -73,7 +73,7 @@ public final class HTMLUtils
         /**
          * Regex to recognize a XML Entity.
          */
-        private static final Pattern ENTITY = Pattern.compile("&[a-zA-Z]+;|&#[0-9a-zA-Z]+;");
+        private static final Pattern ENTITY = Pattern.compile("&[a-zA-Z0-9]+;|&#[0-9a-zA-Z]+;");
 
         /**
          * Ampersand character.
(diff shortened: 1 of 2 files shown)


surli:
@tmortagne this is ready. I checked on the HTML4 RFC https://www.w3.org/TR/html401/sgml/entities.html and following wikipedia pages: https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references I think we cover all cases. If you've a better source to ensure the pattern tell me.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
```diff
@@ -73,7 +73,7 @@
         /**
          * Regex to recognize a XML Entity.
          */
-        private static final Pattern ENTITY = Pattern.compile("&[a-zA-Z]+;|&#[0-9a-zA-Z]+;");
+        private static final Pattern ENTITY = Pattern.compile("&[a-zA-Z0-9]+;|&#[0-9a-zA-Z]+;");
```
vmassol: A documentation link to where this is coming from would be great. Isn't there some grammar that exists somewhere?

surli: Didn't find it yet. If you have an idea where to find it I take it.

vmassol: https://www.w3.org/TR/WD-xml-lang-970630#sec4.3 ?

vmassol: So for example `%_` seems valid but you don't seem to have it in your regex...

vmassol: Next question: is there some existing regex or API to check that?

surli: > My point is that we need to check against the grammar to be sure to not forget anything. 

Yes I agree that was part of my question when posting the sources :)

vmassol:
Note: I haven't checked the grammar so I'm trusting you :)

PR #70, opened 2019-01-31
XCOMMONS-1571: HtmlCleaner strips the leading space in value form attributes

surli:
## Issue

https://jira.xwiki.org/browse/XCOMMONS-1571

## Solution
  * Disable trimmed values property in HtmlCleaner

### DIFF
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
@@ -221,6 +221,9 @@ public class DefaultHTMLCleaner implements HTMLCleaner
         // TODO: handle HTML5 correctly (see: https://jira.xwiki.org/browse/XCOMMONS-901)
         defaultProperties.setHtmlVersion(4);
 
+        // Don't trim spaces in attribute values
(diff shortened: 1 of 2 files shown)


xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
```diff
@@ -221,6 +221,9 @@ private CleanerProperties getDefaultCleanerProperties(HTMLCleanerConfiguration c
         // TODO: handle HTML5 correctly (see: https://jira.xwiki.org/browse/XCOMMONS-901)
         defaultProperties.setHtmlVersion(4);
 
+        // Don't trim spaces in attribute values
```
vmassol: Could you explain why in the comment, maybe by giving an example? I guess that trimming is the default so if we go against the default we should explain why so that we remember it in 2 years time ;) Thx

surli: I can add in the comment my test example, if it's what you have in mind.

vmassol: What I have in mind is to document why we need to trim spaces. Thx

PR #71, opened 2019-02-01
XCOMMONS-1571: HtmlCleaner strips the leading space in value form attributes

surli:
* Create a new dedicated TagTransformation to trim all attributes but
input value.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
@@ -1,0 +1,46 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.htmlcleaner;
+
+import org.xwiki.stability.Unstable;
+
+/**
+ * This class allows to create on the fly a new tag transformation to trim leading space.
+ * See {@link TrimAttributeTagTransformation} for more information.
+ *
+ * @version $Id$
+ * @since 11.1RC1
+ */
+@Unstable
+public class TrimAttributeCleanerTransformations extends CleanerTransformations
+{
+    @Override
+    public TagTransformation getTransformation(String tagName) {
+        TagTransformation transfo = super.getTransformation(tagName);
+        if (transfo == null) {
+
+            // we only create the transformation if it doesn't exist yet
+            // and we keep it to avoid creating multiple objects for the same tag over and over.
+            transfo = new TrimAttributeTagTransformation(tagName, tagName, true);
+            this.addTransformation(transfo);
+        }
+        return transfo;
+    }
+}
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
@@ -34,8 +34,7 @@ import org.htmlcleaner.CleanerTransformations;
 import org.htmlcleaner.DoctypeToken;
 import org.htmlcleaner.HtmlCleaner;
 import org.htmlcleaner.TagNode;
-import org.htmlcleaner.TrimAttributeCleanerTransformations;
-import org.htmlcleaner.TrimAttributeTagTransformation;
+import org.htmlcleaner.TagTransformation;
 import org.htmlcleaner.XWikiDOMSerializer;
 import org.w3c.dom.Document;
 import org.xwiki.component.annotation.Component;
@@ -222,8 +221,9 @@ public class DefaultHTMLCleaner implements HTMLCleaner
         // TODO: handle HTML5 correctly (see: https://jira.xwiki.org/browse/XCOMMONS-901)
         defaultProperties.setHtmlVersion(4);
 
-        // We don't trim any attribute value by default, so we can keep the leading space in input value attributes.
-        // But we use a specific TrimAttributeCleanerTransformation in order to trim the space for any other attribute.
+        // We trim values by default for all attributes but the input value attribute.
+        // The only way to currently do that is to switch off this flag, and to create a dedicated TagTransformation.
+        // See TrimAttributeCleanerTransformation for more information.
         defaultProperties.setTrimAttributeValues(false);
 
         return defaultProperties;
@@ -238,33 +238,34 @@ public class DefaultHTMLCleaner implements HTMLCleaner
     {
         CleanerTransformations defaultTransformations = new TrimAttributeCleanerTransformations();
 
-        TrimAttributeTagTransformation tt = new TrimAttributeTagTransformation(HTMLConstants.TAG_B,
+        // note that we do not care here to use a TrimAttributeTagTransformation, since the attributes are not preserved
+        TagTransformation tt = new TagTransformation(HTMLConstants.TAG_B,
             HTMLConstants.TAG_STRONG, false);
         defaultTransformations.addTransformation(tt);
 
-        tt = new TrimAttributeTagTransformation(HTMLConstants.TAG_I, HTMLConstants.TAG_EM, false);
+        tt = new TagTransformation(HTMLConstants.TAG_I, HTMLConstants.TAG_EM, false);
         defaultTransformations.addTransformation(tt);
 
-        tt = new TrimAttributeTagTransformation(HTMLConstants.TAG_U, HTMLConstants.TAG_INS, false);
+        tt = new TagTransformation(HTMLConstants.TAG_U, HTMLConstants.TAG_INS, false);
         defaultTransformations.addTransformation(tt);
 
-        tt = new TrimAttributeTagTransformation(HTMLConstants.TAG_S, HTMLConstants.TAG_DEL, false);
+        tt = new TagTransformation(HTMLConstants.TAG_S, HTMLConstants.TAG_DEL, false);
         defaultTransformations.addTransformation(tt);
 
-        tt = new TrimAttributeTagTransformation(HTMLConstants.TAG_STRIKE, HTMLConstants.TAG_DEL, false);
+        tt = new TagTransformation(HTMLConstants.TAG_STRIKE, HTMLConstants.TAG_DEL, false);
         defaultTransformations.addTransformation(tt);
 
-        tt = new TrimAttributeTagTransformation(HTMLConstants.TAG_CENTER, HTMLConstants.TAG_P, false);
+        tt = new TagTransformation(HTMLConstants.TAG_CENTER, HTMLConstants.TAG_P, false);
         tt.addAttributeTransformation(HTMLConstants.ATTRIBUTE_STYLE, "text-align:center");
         defaultTransformations.addTransformation(tt);
 
         String restricted = configuration.getParameters().get(HTMLCleanerConfiguration.RESTRICTED);
         if ("true".equalsIgnoreCase(restricted)) {
 
-            tt = new TrimAttributeTagTransformation(HTMLConstants.TAG_SCRIPT, HTMLConstants.TAG_PRE, false);
+            tt = new TagTransformation(HTMLConstants.TAG_SCRIPT, HTMLConstants.TAG_PRE, false);
             defaultTransformations.addTransformation(tt);
 
-            tt = new TrimAttributeTagTransformation(HTMLConstants.TAG_STYLE, HTMLConstants.TAG_PRE, false);
+            tt = new TagTransformation(HTMLConstants.TAG_STYLE, HTMLConstants.TAG_PRE, false);
             defaultTransformations.addTransformation(tt);
         }
 
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
@@ -222,6 +222,9 @@ public class DefaultHTMLCleaner implements HTMLCleaner
         defaultProperties.setHtmlVersion(4);
 
         // Don't trim spaces in attribute values
+        // We don't want input such as: <input type="hidden" value="  foo" />
+        // to be modified for: value="foo"
+        // if such case happen, the value should remain the same.
         defaultProperties.setTrimAttributeValues(false);
 
         return defaultProperties;
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
@@ -34,7 +34,8 @@ import org.htmlcleaner.CleanerTransformations;
 import org.htmlcleaner.DoctypeToken;
 import org.htmlcleaner.HtmlCleaner;
 import org.htmlcleaner.TagNode;
-import org.htmlcleaner.TagTransformation;
+import org.htmlcleaner.TrimAttributeCleanerTransformations;
+import org.htmlcleaner.TrimAttributeTagTransformation;
 import org.htmlcleaner.XWikiDOMSerializer;
 import org.w3c.dom.Document;
 import org.xwiki.component.annotation.Component;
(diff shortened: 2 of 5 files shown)


surli:
@vmassol I improved a bit the comment

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
```diff
@@ -222,6 +222,9 @@ private CleanerProperties getDefaultCleanerProperties(HTMLCleanerConfiguration c
         defaultProperties.setHtmlVersion(4);
 
         // Don't trim spaces in attribute values
+        // We don't want input such as: <input type="hidden" value="  foo" />
+        // to be modified for: value="foo"
+        // if such case happen, the value should remain the same.
```
vmassol: Thanks but you still don't explain why. The comment you put is not needed since that's what 'setTrimmedAttributeValues(false)` is about. You're just explaining WHAT the code does but not WHY :)

I was asking to explain why we want `value = "  foo"` instead of `value = "foo"` since it's far from obvious and it's not the default!

vmassol: Sorry but I still don't see why we wouldn't want spaces to be trimmed in attribute values. Could you explain? I guess that if there was a JIRA issue then there's a reason why this JIRA issue was created.

vmassol: Also, if there's a good reason to not trim spaces, we might want to raise an issue for htmlcleaner to ask them to change the default and explain why it's not a good default. Unless the reason is specific to XWiki ofc.

vmassol: > So again, IMO it should only concern form value attribute. So maybe I should create a filter to only clean those "value" attribute.

Ok I understand now (even though I don't know the exact use case, I can understand that we'd want , in some cases, to keep exactly what the use has entered in a form).

Now the change above removes the trim for all attributes. I wonder if this can cause some troubles in some part of our code where we had leading or trailing spaces that were fixed by the html cleaner. Also if users use our `{{html}}` macro the HTML Cleaner is executed and this if they were using HTML with attribute values with leading/trailing spaces maybe it was good that our HTML cleaner was "fixing" them for their use cases and maybe their use case will fail after the change?

So indeed, it could be safer to have a filter for that to limit the trimming for form elements.

Now maybe this is overkill, I'm not sure.

In any case, for the comment, I'd say the we need HTML Form elements to not have their attributes trimmed because we want to keep the user-entered content intact (as it was entered) when doing further processing on the values.

surli: > Now maybe this is overkill, I'm not sure.

The only usecase I can see where we might break something, is for example in case of an html containing classes with leading space and a javascript checking exactly the class attribute value to do something. 

So it's clearly an edge usecase, but it might exist indeed. 

> So indeed, it could be safer to have a filter for that to limit the trimming for form elements.

I'll reopen the issue then.

surli:
@vmassol so it's ready. All tests passing: it only keeps spaces for input value attributes.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
```diff
@@ -0,0 +1,46 @@
...
+import org.xwiki.stability.Unstable;
+
+/**
+ * This class allows to create on the fly a new tag transformation to trim leading space.
+ * See {@link TrimAttributeTagTransformation} for more information.
+ *
+ * @version $Id$
+ * @since 11.1RC1
+ */
+@Unstable
```
vmassol: Do we need this class to be public?

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
```diff
@@ -0,0 +1,46 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.htmlcleaner;
```
vmassol: hmm so this is not something meant to be extended. We definitely need a javadoc comment below to explain why we're doing this and what is the good future solution, and reference an htmlcleaner issue to follow up and when fixed explain to remove this class.

vmassol: We should also mention that this class is not meant to be public

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @version $Id$
+ * @since 11.1RC1
+ */
+@Unstable
+public class TrimAttributeCleanerTransformations extends CleanerTransformations
+{
+    @Override
+    public TagTransformation getTransformation(String tagName) {
+        TagTransformation transfo = super.getTransformation(tagName);
```
vmassol: Our best practice is not use shorthand names. Bette use `transformation`.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
tmortagne: Definitely super weird to have such a package and it deserve a big comment.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
```diff
@@ -221,10 +222,8 @@ private CleanerProperties getDefaultCleanerProperties(HTMLCleanerConfiguration c
         // TODO: handle HTML5 correctly (see: https://jira.xwiki.org/browse/XCOMMONS-901)
         defaultProperties.setHtmlVersion(4);
 
-        // Don't trim spaces in attribute values
-        // We don't want input such as: <input type="hidden" value="  foo" />
-        // to be modified for: value="foo"
-        // if such case happen, the value should remain the same.
+        // We don't trim any attribute value by default, so we can keep the leading space in input value attributes.
```
vmassol: > We don't trim any attribute value by default

I think it's the opposite. We want to trim attribute values by default except for HTML INPUT tag attributes. Right?

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * This class allows to create on the fly a new tag transformation to trim leading space.
+ * See {@link TrimAttributeTagTransformation} for more information.
+ *
+ * @version $Id$
+ * @since 11.1RC1
+ */
+@Unstable
+public class TrimAttributeCleanerTransformations extends CleanerTransformations
```
vmassol: I would call this class XWikiCleanerTransformations instead.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
```diff
@@ -237,34 +236,35 @@ private CleanerProperties getDefaultCleanerProperties(HTMLCleanerConfiguration c
      */
     private CleanerTransformations getDefaultCleanerTransformations(HTMLCleanerConfiguration configuration)
     {
-        CleanerTransformations defaultTransformations = new CleanerTransformations();
+        CleanerTransformations defaultTransformations = new TrimAttributeCleanerTransformations();
```
vmassol: Name sounds weird, see above for the suggested class name change.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
surli: I'll add a comment but FTR the package was already existing in our own code for XWikiDOMSerializer.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
```diff
@@ -237,34 +236,35 @@ private CleanerProperties getDefaultCleanerProperties(HTMLCleanerConfiguration c
...
 
-        TagTransformation tt = new TagTransformation(HTMLConstants.TAG_B, HTMLConstants.TAG_STRONG, false);
+        TrimAttributeTagTransformation tt = new TrimAttributeTagTransformation(HTMLConstants.TAG_B,
+            HTMLConstants.TAG_STRONG, false);
         defaultTransformations.addTransformation(tt);
 
-        tt = new TagTransformation(HTMLConstants.TAG_I, HTMLConstants.TAG_EM, false);
+        tt = new TrimAttributeTagTransformation(HTMLConstants.TAG_I, HTMLConstants.TAG_EM, false);
```
vmassol: Why do we need to use our own transformation class instead of the HTMLCleaner one? Don't we need our custom one only for INPUT elements?

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -409,10 +409,12 @@ public void verifyEntitiesAreNotBroken()
...
     public void verifyFormValuesAreNotTrimmed()
     {
+        assertHTML("<p><input type=\"hidden\" value=\"  fff\" /></p>", "<input type=\"hidden\" value=\"  fff\" />");
         assertHTML("<p><input type=\"hidden\" value=\"foo\" /></p>", "<input type=\"hidden\" value=\"foo\" />");
         assertHTML("<p><input type=\"hidden\" value=\"foo bar\" /></p>", "<input type=\"hidden\" value=\"foo bar\" />");
-        assertHTML("<p><input type=\"hidden\" value=\"  fff\" /></p>", "<input type=\"hidden\" value=\"  fff\" />");
-        assertHTML("<p><input class=\"  fff\" type=\"hidden\" /></p>", "<input type=\"hidden\" class=\"  fff\" />");
+        assertHTML("<p><input class=\"fff\" type=\"hidden\" /></p>", "<input type=\"hidden\" class=\"  fff\" />");
+        assertHTML("<p><input class=\"foo bar\" type=\"hidden\" value=\" foo bar  \" /></p>",
+            "<input type=\"hidden   \" value=\" foo bar  \" class=\" foo bar  \"/>");
```
vmassol: Do we have another test where we verify that attributes are trimmed by default (when they're not INPUT tag elements)?

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
surli: Yep, that is what we want. But here I explain why we deactivate the flag: we don't let HtmlCleaner trim the attribute. We do it ourselves through the transformation.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
tmortagne: Doesn't make it right IMO :)

If we don't need this package for technical reason we should not use it (so from what I see it was wrong in case of XWikiDOMSerializer).

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
surli: I prefer to see in the name what it's meant for. I can always prefix it with XWiki: XWikiTrimAttributeCleanerTransformation

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
vmassol: > (so from what I see it was wrong in case of XWikiDOMSerializer).

It was necessary ofc, or it wouldn't have been used... It's because we needed to call package protected methods.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
surli: No we don't. As I mentioned: we skip trimming the attributes, so we need to transform ALL the attributes, except the input one.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
surli: Actually in this test I'm checking the behaviour on different kind of attributes. But sure I can add one assert with a div if you like.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
vmassol: ok I understand now but it wasn't very clear. I thought you were explaining what we want (which is what I would expect in such a comment). Explain what we want and then explain how to get there.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
tmortagne: > we needed to call package protected methods

Does not seems to be needed anymore (we don't use the method anymore and it's now public anyway).

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
vmassol: No it's fine, it's because I didn't understand that the logic was reversed.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
vmassol: Then it should be moved indeed.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
vmassol: Maybe I'm not seeing clear but I see only `<input>` tests.

vmassol: And the test says `verifyFormValuesAreNotTrimmed`.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
surli: It's still needed for createSubNodes call in XWikiDOMSerializer and it's definitely needed for the new files I added.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
surli: Yep input but you can see that I'm checking the class and type attributes too.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
tmortagne: > it's definitely needed for the new files I added.

Should be documented then.

surli: Just done.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
surli: I updated the comment.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
surli: Changed that: actually we don't need it since the attribute are not preserved in those.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
surli: Added an assert and changed the test name.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/TrimAttributeCleanerTransformations.java
surli: Fixed

PR #76, opened 2019-08-23
[misc] exclude commons-beanutils dependency from checkstyle

romani:
to avoid conflict with their too recent version

it will allow to merge - https://github.com/checkstyle/checkstyle/pull/7000 (version bump to resolve security issue in checkstyle)

### DIFF
--- a/xwiki-commons-tools/xwiki-commons-tool-verification-resources/pom.xml
+++ b/xwiki-commons-tools/xwiki-commons-tool-verification-resources/pom.xml
@@ -55,6 +55,10 @@
           <groupId>com.google.guava</groupId>
           <artifactId>guava</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>commons-beanutils</groupId>
+          <artifactId>commons-beanutils</artifactId>
+        </exclusion>
       </exclusions>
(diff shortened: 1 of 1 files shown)


tmortagne:
Would make more sense to upgrade beanutils actually. Strangely I did not get any notification for this new beanutils version, otherwise I would have upgraded it.

Taking care of it today. See https://jira.xwiki.org/browse/XCOMMONS-1712.

romani:
It would be probably good to exclude it, if any new version of this library happen it will result in same issue on our side, and will bother you again. Same as we have guava library.

tmortagne:
We tend to upgrade this kind of libraries as soon as they are released but if you are sure it's not needed at build time then sure why not.

romani:
We use only few items from it.
Ok, let's think about it, on next update, it  does not release that often.

tmortagne:
OK you will be able to tell me "I told you so" ;)

romani:
> "I told you so"

@tmortagne , fyi - https://jira.xwiki.org/browse/XCOMMONS-1815

tmortagne:
Not fair, was released just 2 days ago but OK fine :)

PR #79, opened 2019-09-23
Use maven instead of gradle on LGTM

samlanning:


vmassol:
Thanks @samlanning :)

PR #84, opened 2019-12-30
config: remove deprecated properties from checkstyle's JavadocMethod

romani:
https://checkstyle.org/releasenotes.html#Release_8.28

reason: https://github.com/checkstyle/checkstyle/issues/7329

checkstyle project need this update to continue to keep xwiki in checkstyle's CI verification process.

### DIFF
--- a/xwiki-commons-tools/xwiki-commons-tool-verification-resources/src/main/resources/checkstyle.xml
+++ b/xwiki-commons-tools/xwiki-commons-tool-verification-resources/src/main/resources/checkstyle.xml
@@ -209,8 +209,6 @@
 
     <module name="JavadocMethod">
       <property name="scope" value="public"/>
-      <property name="allowUndeclaredRTE" value="true"/>
-      <property name="suppressLoadErrors" value="true" />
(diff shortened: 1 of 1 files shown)


vmassol:
Thanks @romani ! Since these properties are deprecated what are the new equivalents? Won't we loose checks by just removing them?

vmassol:
Answering to myself:

> Migration Notes:
properties allowMissingThrowsTags, allowUndeclaredRTE and allowThrowsTagsForSubclasses, logLoadErrors, suppressLoadErrors should be just removed from configs.
Behavior will change as there will be less violations in code, such properties did not work well so their removal should be ok, as there will be less false-positives. Throws will not be validated by default !!
If you need to validate @throws please enable property validateThrows but please read documentation as there some limitations.

vmassol:
So we need to decide if we want `validateThrows`.

vmassol:
@tmortagne what about my comments/questions? :)

tmortagne:
I think the javadoclint build is validating this kind of thing already.

PR #92, opened 2020-04-01
XWIKI-5950: Create an ApplicationEvent interface as a parent for the different ApplicationEvents

Haxsen:
Hi, this is my proposed solution.
Will be glad to clear my mistakes by your reviews.
Thanks.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
+++ b/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
@@ -22,7 +22,7 @@ package org.xwiki.observation.event;
 /**
  * All Event types must implement this interface.
  *
- * @version $id$
+ * @version $id$ (return the application event id)
  */
 public interface ApplicationEvent
 {
--- a/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
+++ b/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
@@ -20,7 +20,7 @@
 package org.xwiki.observation.event;
 
 /**
- * All Event types must implement this interface.
+ * All ApplicationEvent types must implement this interface.
  *
  * @version $id$ (return the application event id)
  */
--- a/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
+++ b/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
@@ -22,7 +22,7 @@ package org.xwiki.observation.event;
 /**
  * All ApplicationEvent types must implement this interface.
  *
- * @version $id$ (return the application event id)
+ * @version $id$
  */
 public interface ApplicationEvent extends Event
 {
--- a/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
+++ b/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
@@ -1,0 +1,42 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.observation.event;
+
+/**
+ * All Event types must implement this interface.
+ *
+ * @version return the event id (When the Application starts or ends)
+ */
+public interface ApplicationEvent
+{
+    /**
+     * Creates a non-generic parent interface for ApplicationStartedEvent and ApplicationStoppedEvent
+     * Compares two events to see if they <em>match</em>, meaning that a listener that registered to receive
+     * notifications <em>like</em> <code>referenceEvent</code> will be notified of any occuring event for which
+     * <code>referenceEvent.matches(occuringEvent)</code> will return <code>true</code>. The matching algorithm depends
+     * on the event implementation. For example for Document events two events match if they implement the same event
+     * class and if their {@link org.xwiki.observation.event.filter.EventFilter} match. Note that the implementation is
+     * left open in order to cater for all the possible ways this Observation component can be used.
+     *
+     * @param otherEvent the occuring event matched against the current object
+     * @return <code>true</code> if the passed event matches this event, <code>false</code> otherwise.
+     */
+    boolean matches(Object otherEvent);
+}
--- a/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
+++ b/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
@@ -22,7 +22,7 @@ package org.xwiki.observation.event;
 /**
  * All Event types must implement this interface.
  *
- * @version return the event id (When the Application starts or ends)
+ * @version $id$
  */
 public interface ApplicationEvent
 {
--- a/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
+++ b/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
@@ -22,7 +22,7 @@ package org.xwiki.observation.event;
 /**
  * All ApplicationEvent types must implement this interface.
  *
- * @version $id$
+ * @version $Id$
  */
 public interface ApplicationEvent extends Event
 {
--- a/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
+++ b/xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
@@ -24,7 +24,7 @@ package org.xwiki.observation.event;
  *
  * @version $id$ (return the application event id)
  */
-public interface ApplicationEvent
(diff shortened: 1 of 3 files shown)


xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.observation.event;
+
+/**
+ * All Event types must implement this interface.
```
vmassol, line 23: Why?

Haxsen: whoopsie, going to change that to ApplicationEvent.

vmassol:
I don't understand why what you did is useful. Could you elaborate?

The issue doesn't seem to explain why it's useful and why it should be done

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
```diff
@@ -20,7 +20,7 @@
 package org.xwiki.observation.event;
 
 /**
- * All Event types must implement this interface.
+ * All ApplicationEvent types must implement this interface.
```
vmassol, line 23: Why?

vmassol: also what is an `ApplicationEvent`?

Haxsen:
`Event` is more like a generic name for Starting and Stopping of an `ApplicationEvent` so `ApplicationEvent` seems more suitable for `ApplicationStartedEvent` and `ApplicationStoppedEvent`.

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
vmassol: My point is that I think you're doing a PR without understanding the code and the underlying things being done.

vmassol:
> so ApplicationEvent seems more suitable for ApplicationStartedEvent and ApplicationStoppedEvent.

WDYM by suitable? Based on what criteria?

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
Haxsen: Since they should let know that they are child of this interface which is a parent.
I might be wrong, can you correct me?

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
```diff
@@ -0,0 +1,42 @@
...
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.observation.event;
+
+/**
+ * All ApplicationEvent types must implement this interface.
+ *
+ * @version $id$ (return the application event id)
+ */
+public interface ApplicationEvent
```
tmortagne, line 27: As I told you in the issue, you are not supposed to completely replace Event. The point is to have a common interface for ApplicationStartedEvent and ApplicationStoppedEvent but as you saw yourself in the Event javadoc all events have to implement it so ApplicationEvent should extend Event (otherwise it not an event).

Haxsen: Oh alright, yes I forgot to extend it. Doing it.

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
```diff
@@ -0,0 +1,42 @@
...
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.observation.event;
+
+/**
+ * All ApplicationEvent types must implement this interface.
+ *
+ * @version $id$ (return the application event id)
```
vmassol, line 25: ??? :)

Haxsen:
> > so ApplicationEvent seems more suitable for ApplicationStartedEvent and ApplicationStoppedEvent.
> 
> WDYM by suitable? Based on what criteria?

I think it gives a better structure of the classes and interfaces.

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
vmassol: BTW I think this shows you haven't executed the build or tested the change since that would clearly break XWiki....

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
Haxsen: Yes I am confused about this.. I think I should remove the braces sentence.

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
Haxsen: @vmassol  I tested using `mvn clean test` , can you suggest better way? I mean I can build but this is not XWiki-Platform where I can build and test the real application.

vmassol: ok it means there's no tests in this module if the build passed. What you can do to check that a change works is to build the module and copy the JAR in an xwiki instance in WEB-INF/lib and restart.

Haxsen: @vmassol that's helpful. Thanks.

vmassol:
> I think it gives a better structure of the classes and interfaces.

Ok, sure, but there needs to be a usage of it and your PR doesn't show a usage of the new structure. Otherwise it's just YAGNI (https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
```diff
@@ -0,0 +1,42 @@
...
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.observation.event;
+
+/**
+ * All ApplicationEvent types must implement this interface.
+ *
+ * @version $id$
+ */
```
tmortagne, line 26: You should add a

```
@since 12.3RC1
```

in the Javadoc to indicate since when this interface exist.

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
```diff
@@ -0,0 +1,42 @@
...
+     * notifications <em>like</em> <code>referenceEvent</code> will be notified of any occuring event for which
+     * <code>referenceEvent.matches(occuringEvent)</code> will return <code>true</code>. The matching algorithm depends
+     * on the event implementation. For example for Document events two events match if they implement the same event
+     * class and if their {@link org.xwiki.observation.event.filter.EventFilter} match. Note that the implementation is
+     * left open in order to cater for all the possible ways this Observation component can be used.
+     *
+     * @param otherEvent the occuring event matched against the current object
+     * @return <code>true</code> if the passed event matches this event, <code>false</code> otherwise.
+     */
+    boolean matches(Object otherEvent);
```
tmortagne, line 41: This is already part of Event.

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
```diff
@@ -0,0 +1,42 @@
...
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.observation.event;
+
+/**
+ * All ApplicationEvent types must implement this interface.
+ *
+ * @version $id$
```
vmassol, line 25: Note you need a `@since` and an `@Unstable` annotation. See https://dev.xwiki.org/xwiki/bin/view/Main/WebHome which contain all our dev rules. In this specific case, it's https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HBackwardCompatibility

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationStartedEvent.java
```diff
@@ -25,7 +25,7 @@
  * @version $Id$
  * @since 2.0M1
  */
-public class ApplicationStartedEvent implements Event
+public class ApplicationStartedEvent implements ApplicationEvent
```
vmassol, line 28: Open question: does this break revapi (ie the backward compatibility check)?

vmassol:
What about `ApplicationReadyEvent` ? is that an Application Event too?

vmassol:
I still don't understand the need for this issue. I would close it as won't fix unless someone can come up with a use case where we can use it right now.

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
Haxsen: Yep, I think this issue only creates extra unnecessary interface. Should I drop this issue?

Haxsen:
I am closing this PR as the issue doesn't seem to have a use case.
What I learned:
1. I should check reverse compatibility.
2. If I add a new file, use @since <version> to show where it was introduced.
3. I can test XWiki-Commons modified jar files by building and replacing in the XWiki instance in WEB-INF/lib.

Thanks @vmassol  and @tmortagne for the valuable review.

xwiki-commons-core/xwiki-commons-observation/xwiki-commons-observation-api/src/main/java/org/xwiki/observation/event/ApplicationEvent.java
Haxsen: Thanks @tmortagne .

PR #93, opened 2020-04-02
XCOMMONS-1740: Upgrade to HTML Cleaner 2.23

surli:
* Upgrade to 2.23
  * Fix XWikiDOMSerializer to use the right flag to not escape XML in
    the produced Document
  * Fix a bug in HTMLUtils that prevented the < > to be properly escaped
    during serialization
  * Refactor HtmlCleanerTest to ensure that the content of the Document
    is what expected before serialization. And change some expected HTML
results.
  * Introduce a new configuration in HtmlCleanerConfiguration to
    translate special HTML entities, to be able to use them in an XML
Parser
  * Use the new configuration in UnifiedHTMLDiffManager
  * Fix the failing test and clean a bit

### DIFF
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/XWikiDOMSerializer.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/XWikiDOMSerializer.java
@@ -133,7 +133,6 @@ public class XWikiDOMSerializer extends DomSerializer
             //
             if (!props.isAllowInvalidAttributeNames()) {
                 attrName = Utils.sanitizeXmlIdentifier(attrName, props.getInvalidXmlAttributeNamePrefix());
-                //attrName = Utils.sanitizeXmlAttributeName(attrName, props.getInvalidXmlAttributeNamePrefix());
             }
 
             if (attrName != null && (Utils.isValidXmlIdentifier(attrName) || props.isAllowInvalidAttributeNames())) {
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/XWikiDOMSerializer.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/XWikiDOMSerializer.java
@@ -58,7 +58,8 @@ public class XWikiDOMSerializer extends DomSerializer
      */
     public XWikiDOMSerializer(CleanerProperties props)
     {
-        super(props);
+        // We don't want the XML to be escaped in the produced Document.
+        super(props, false);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


surli:
Note: I ran `mvn clean install -Pquality` on commons and rendering locally and everything's passing locally. I won't run all platform locally, but I will try to identify the places where HtmlCleaner was used to run those modules and fix the hack we performed. I already identified at least one hack in platform-office-importer which was the root cause of the bug report last year.

surli:
So I looked for occurences of HtmlCleaner in platform and the only occurences I found were located in xwiki-platform-office-importer. I fixed the location I knew where we were manually transforming filenames because of the HtmlCleaner bug and ran tests on office-importer and office-test-docker: all green. 
Here's the related commit: https://github.com/xwiki/xwiki-platform/commit/321dbfb7da989f948be87cef1b3e8597973eca22

Note that I also manually rebuilt a distrib and perform the import of the file from https://jira.xwiki.org/browse/XWIKI-15930 to ensure it's still fixed after my change: all good. 

So for me it's all ready.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/XWikiDOMSerializer.java
```diff
@@ -131,7 +132,7 @@ private Document createDocument(DocumentBuilder builder, TagNode rootNode)
             // Fix any invalid attribute names
             //
             if (!props.isAllowInvalidAttributeNames()) {
-                attrName = Utils.sanitizeXmlAttributeName(attrName, props.getInvalidXmlAttributeNamePrefix());
+                attrName = Utils.sanitizeXmlIdentifier(attrName, props.getInvalidXmlAttributeNamePrefix());
```
mflorea, line 135: This probably needs a comment explaining why the dedicated ``sanitizeXmlAttributeName`` is not used.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
```diff
@@ -134,7 +134,7 @@ public String escapeElementEntities(String text)
                 }
             } else {
                 result = escapeAmpersand(text);
-                StringUtils.replaceEach(text, REPLACE_ELEMENTS_SEARCH, REPLACE_ELEMENTS_RESULT);
+                result = StringUtils.replaceEach(result, REPLACE_ELEMENTS_SEARCH, REPLACE_ELEMENTS_RESULT);
```
mflorea, line 137: This looks like a bug fix unrelated to the HTMLCleaner upgrade.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/XWikiDOMSerializer.java
surli: it's actually a rename inside HtmlCleaner: the method `Utils.sanitizeXmlAttributeName` does not exist anymore.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
surli: It's not directly related to the upgrade, but it's related to the fix of escaped values. I put everything on the same basket here.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/htmlcleaner/XWikiDOMSerializer.java
mflorea: OK.

PR #94, opened 2020-04-30
XCOMMONS-1938: Upgrade to HtmlCleaner 2.24

surli:
* Keep as much as possible the entities parsed in the Document
  * Since HtmlCleaner doesn't allow to easily parse unicode characters,
    keep them translated with their character reference
  * Only unescape the &amp; for character references in HtmlUtils
  * Remove previous exception made in UnifiedHTMLDiffManager since it's
    not needed anymore.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
@@ -98,6 +98,7 @@ public class DefaultHTMLCleanerTest
     @Test
     void specialCharacters()
     {
+        // The blank space is not a standard space, but a non-breaking space.
         assertHTML("<p>\"&amp;**notbold**&lt;notag&gt; </p>",
             "<p>&quot;&amp;**notbold**&lt;notag&gt;&nbsp;</p>");
         assertHTML("<p>\"&amp;</p>", "<p>\"&</p>");
--- a/xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
@@ -535,6 +535,7 @@ public class DefaultHTMLCleanerTest
         // HtmlCleaner currently doesn't handle properly unicode characters: asking it to recognize them
         // involves that all entities will be escaped during the parsing and that's not what we want. So we
         // keep them encoded.
+        // See https://sourceforge.net/p/htmlcleaner/bugs/221/
         textContent =
             document.getElementsByTagName("p").item(0).getTextContent();
         assertEquals("&#8;", textContent);
@@ -618,5 +619,50 @@ public class DefaultHTMLCleanerTest
 
         assertHTML("<div foo=\"&amp;quot;\">content</div>",
             "<div foo=\"&amp;quot;\">content</div>");
+    }
+
+    @Test
+    @Disabled("See https://sourceforge.net/p/htmlcleaner/bugs/221/")
+    public void parseWithUnicodeChars() throws Exception
+    {
+        CleanerProperties cleanerProperties = new CleanerProperties();
+
+        cleanerProperties.setDeserializeEntities(true);
+        cleanerProperties.setRecognizeUnicodeChars(true);
+        cleanerProperties.setTranslateSpecialEntities(false);
+
+        HtmlCleaner cleaner = new HtmlCleaner(cleanerProperties);
+        TagNode tagNode = cleaner.clean("<?xml version = \"1.0\"?>"
+            + "<div foo=\"&#169;\">&#169;</div>"
+            + "<div foo=\"baz&gt;buz\">baz&gt;buz</div>"
+            + "<div foo=\"baz&buz\">baz&buz</div>");
+        List<? extends TagNode> divList = tagNode.getElementListByName("div", true);
+        assertEquals(3, divList.size());
+
+        assertEquals("©", divList.get(0).getText().toString());
+        assertEquals("©", divList.get(0).getAttributeByName("foo"));
+
+        assertEquals("baz>buz", divList.get(1).getText().toString());
+        assertEquals("baz>buz", divList.get(1).getAttributeByName("foo"));
+
+        assertEquals("baz&buz", divList.get(2).getText().toString());
+        assertEquals("baz&buz", divList.get(2).getAttributeByName("foo"));
+
+        DomSerializer domSerializer = new DomSerializer(cleanerProperties, false);
+        Document document = domSerializer.createDOM(tagNode);
+
+        NodeList nodeList = document.getElementsByTagName("div");
+        assertEquals(3, nodeList.getLength());
+        assertEquals("©", nodeList.item(0).getTextContent());
+        assertEquals("©", nodeList.item(0).getAttributes().getNamedItem("foo").getTextContent());
+
+        assertEquals("baz>buz", nodeList.item(1).getAttributes().getNamedItem("foo").getTextContent());
+
+        assertEquals("baz&buz", nodeList.item(2).getAttributes().getNamedItem("foo").getTextContent());
+
+        // BUG: This is failing with baz&gt;buz
+        assertEquals("baz>buz", nodeList.item(1).getTextContent());
+        // BUG: This is failing with baz&amp;buz
+        assertEquals("baz&buz", nodeList.item(2).getTextContent());
     }
 }
--- a/xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
@@ -26,9 +26,14 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.htmlcleaner.CleanerProperties;
+import org.htmlcleaner.DomSerializer;
+import org.htmlcleaner.HtmlCleaner;
+import org.htmlcleaner.TagNode;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.w3c.dom.Document;
+import org.w3c.dom.NodeList;
 import org.xwiki.component.manager.ComponentManager;
 import org.xwiki.test.annotation.ComponentList;
 import org.xwiki.test.junit5.mockito.ComponentTest;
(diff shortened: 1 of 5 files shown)


xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -489,18 +490,18 @@ public void escapeHTMLCharsInAttributes() throws Exception
 
         String textContent =
             document.getElementsByTagName("div").item(0).getAttributes().getNamedItem("foo").getTextContent();
-        assertEquals("aaa\"bbb&ccc>ddd<eee'fff", textContent);
+        assertEquals("aaa&quot;bbb&amp;ccc&gt;ddd&lt;eee&apos;fff", textContent);
```
tmortagne, line 493: I don't understand. Isn't this the opposite of what we want ? We want to have real content in the Document, not escaped content which only make sense in the serialized xml.

surli: > We want to have real content in the Document

Well the HTML input here was:  `"<div foo=\"aaa&quot;bbb&amp;ccc&gt;ddd&lt;eee&apos;fff\">content</div>";` so it is actually the real content.

mflorea: @surli I don't think it's correct. You can try it in the browser:

```
> jQuery('#document-title').html('<h1 foo="aaa&quot;bbb&amp;ccc&gt;ddd&lt;eee&apos;fff">Test</h1>')
> jQuery('#document-title').find('h1').attr('foo')
aaa"bbb&ccc>ddd<eee'fff
```
The special characters are escaped when the HTML is serialized:

> String htmlInput = "<div foo=\"aaa&quot;bbb&amp;ccc&gt;ddd&lt;eee&apos;fff\">content</div>";

but they are not escaped when you retrieve the attribute value from the DOM.

tmortagne: > Well the HTML input here was: "<div foo=\"aaa&quot;bbb&amp;ccc&gt;ddd&lt;eee&apos;fff\">content</div>"; so it is actually the real content.

No it's not. The real content is the unescaped content. You need to escape to serialize the XML but it does not make any sense to have &quot; in the Document.

surli: I'm not sure I understand, I'm just telling you what is given as input of the test, see https://github.com/xwiki/xwiki-commons/pull/94/files#diff-662b1d8baf4677911842867688c409bfR488

tmortagne: We did look at the test input and the test input is XML (XHTML) content, when you parse XML content you are supposed to unescape it so that you find unescaped value in the Document. "&quot;" is the XML serialized version of `"` (i.e. the real value) so you are supposed to find `"` in the Document after parsing it since the point of Document is a Java representation of a XML DOM with the real values in it.

surli: Ok so you are saying that my other test (which was actually the bug I asked Scott to fix) was wrong too see https://github.com/xwiki/xwiki-commons/pull/94/files#diff-662b1d8baf4677911842867688c409bfR575

tmortagne: https://github.com/xwiki/xwiki-commons/pull/94/files#diff-662b1d8baf4677911842867688c409bfR575 is definitely wrong yes. The value of the foo attribute is `&quot;`.

We discussed this many times and the initial issue that leaded to all the work around html cleaner was that escaped html was kept escaped in the Document which did not made any sense so we definitely don't want to come back to it... And honestly really I don't understand why we keep having this kind of conversations since for me it's very simple: escaping is XML syntax exactly like '<' is so it only make sense in the serialized XML String. The DOM does not contain XML syntax since the point of this syntax is to describe the DOM (exactly like you don't end up with `"<div>"` when you ask for the element name).

surli: > And honestly really I don't understand why we keep having this kind of conversations

Sorry for being such a pain on that one. I think there has been many misunderstanding because of two things:
  1. We don't just have one parse and one serializer, we have one parser that generates a `TagNode`, then a first serializer that generates a `Document` and then a second serializer that generates the html. So there was lot of confusion on which serializer we were talking about
  2. in DefaultHtmlCleanerTest we are actually testing two things: DefaultHtmlCleaner and HTMLUtils (the html serializer) which really does not help...

So I continued to investigate on it by changing https://github.com/xwiki/xwiki-commons/pull/94/files#diff-662b1d8baf4677911842867688c409bfR575 to have:
```
document = this.cleaner.clean(new StringReader("<div foo=\"&amp;quot;\">&amp;quot;</div>"));
nodeList = document.getElementsByTagName("div");
assertEquals(1, nodeList.getLength());
assertEquals("&quot;", nodeList.item(0).getTextContent());
assertEquals("&quot;",nodeList.item(0).getAttributes().getNamedItem("foo").getTextContent());

assertHTML("<div foo=\"&amp;quot;\">content</div>", "<div foo=\"&amp;quot;\">content</div>");
```

I manage to make it work for DefaultHTMLCleaner but not for HTMLUtils: until now in HTMLUtils we were expecting to have the serialized entities from the Document and to not touch them, while now if we encounter a `&quot;` it would mean that we really have a `&amp;quot;`. So I will change this behaviour in HTMLUtils but I'm a bit afraid to break lots of things.

surli: @tmortagne so for info, the changes we perform to have everything deserialized in the Document means that the following test: https://github.com/xwiki/xwiki-commons/blob/master/xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java#L428-L433 is completely broken right now, because we were never expecting HTMLUtils to serialize back the entities. So I need to find a way to do that processing in HTMLUtils.

surli: From what I see the issue is that HtmlUtils is actually using an XMLOutputter, so it doesn't take into account HTML entities at all. Only the XML entities...

surli: Note that AFAICS HtmlCleaner provides an HtmlSerializer which IMO would fit better for HTMLUtils. The issue is that our API for HTMLUtils is taken a Document, where HtmlCleaner takes as input a TagNode. So if we are willing to use it we would need to transform back the Document to TagNode. (back because when we parse the html with HtmlCleaner we obtain a TagNode that we serialize then to Document)

tmortagne: > From what I see the issue is that HtmlUtils is actually using an XMLOutputter, so it doesn't take into account HTML entities at all. Only the XML entities...

The serializer should escape only what needs to be escaped in the context of a specific encoding (usually UTF8 in our case). While the parser need to understand all kind of entities it might encounter to resolve them properly very few of the characters they represent really needed to be escaped in the first place in UTF8 (probably none of them actually).

surli: So you're saying that: 
```
// This is wrong
assertHTML("<p>&Eacute;</p>", "&Eacute;");

// This is right
assertHTML("<p>É</p>", "&Eacute;");
```

If it's the case I don't have anything to do with HTMLUtils indeed, but I'm really afraid that we break lots of assertions everywhere we use some html input/output. First example I have in mind is the diff-displayer tests which is using HTMLUtils but maybe there's other places too.

tmortagne: It's not wrong but IMO it's not a requirement to keep `&Eacute;` as `&Eacute;` when cleaning the content as long as the target encoding supports it.

surli: > It's not wrong but IMO it's not a requirement to keep &Eacute; as &Eacute; when cleaning the content as long as the target encoding supports it.

OK going to change those assertions then, and I'll run the build on commons to see what happens.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -518,16 +530,30 @@ public void controlCharacters() throws Exception
         htmlInput = "<p>&#8;</p>";
         document = this.cleaner.clean(new StringReader(htmlInput));
 
+        // HtmlCleaner currently doesn't handle properly unicode characters: asking it to recognize them
+        // involves that all entities will be escaped during the parsing and that's not what we want. So we
+        // keep them encoded.
         textContent =
             document.getElementsByTagName("p").item(0).getTextContent();
         assertEquals("&#8;", textContent);
```
tmortagne, line 538: I would make more explicit that this behavior is insane with a TODO and an issue on HTMLCleanr side.

surli:
Could be good to have your opinion on last changes @mflorea  I'd like to merge it quickly to have CI feedback for 12.5RC1

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -105,7 +96,7 @@ public void elementExpansion()
     @Test
     public void specialCharacters()
     {
-        assertHTML("<p>&quot;&amp;**notbold**&lt;notag&gt;&nbsp;</p>",
+        assertHTML("<p>\"&amp;**notbold**&lt;notag&gt; </p>",
```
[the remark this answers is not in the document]
surli: Good point, I'll double-check and put the info. Too bad that Github doesn't provide a way to display it :)

PR #96, opened 2020-06-02
XCOMMONS-1959: Allow to define the pool size of a GroupedJob

surli:
* Define a GroupedJobInitializer role whose purpose is to be able to
    specify some default configuration for the pool of GroupedJob
  * Define a GroupedJobInitializerManager in charge of retrieving the
    appropriate GroupedJobInitializer based on a JobGroupPath and to
fallback on parents or default
  * Rewrite JobGroupPathLockTree to use read/write semaphores based on
    the configured pool size
  * Change ThreadPoolExecutor constructors in DefaultJobExecutor to
    allow using a pool of more than 1 threads and to allow configuring
the keepAliveTime for the threads
  * Provide some new tests to ensures the threads are operating
    correctly with a pool > 1 thread
  * Create a dedicated class for ReadWriteSemaphore and improve a bit
    the semaphore implementation with a lock on readCounter and by
taking into account the volatile for counters atomic access
  * Change default value for JobGroupInitializer cache size
  * Invalidate the cache in DefaultGroupedJobInitializerManager if a
    component is removed.
  * Makes JobGroupPath explicitely serializable.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/JobManagerConfiguration.java
+++ b/xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/JobManagerConfiguration.java
@@ -52,7 +52,7 @@ public interface JobManagerConfiguration
     @Unstable
     default int getGroupedJobInitializerCacheSize()
     {
-        return 10;
+        return 100;
     }
 
     /**
--- a/xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/JobManagerConfiguration.java
+++ b/xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/JobManagerConfiguration.java
@@ -20,8 +20,10 @@
 package org.xwiki.job;
 
 import java.io.File;
+import java.util.concurrent.TimeUnit;
 
 import org.xwiki.component.annotation.Role;
+import org.xwiki.stability.Unstable;
 
 /**
  * Some job manager related configuration.
@@ -42,4 +44,36 @@ public interface JobManagerConfiguration
      * @since 7.2M2
      */
     int getJobStatusCacheSize();
+
+    /**
+     * @return the number of {@link GroupedJobInitializer} to keep in cache.
+     * @since 12.5RC1
+     */
+    @Unstable
+    default int getGroupedJobInitializerCacheSize()
+    {
+        return 10;
+    }
+
+    /**
+     * @return the duration in milliseconds for the thread keep alive in our single job thread executor.
+     * @see java.util.concurrent.ThreadPoolExecutor#setKeepAliveTime(long, TimeUnit).
+     * @since 12.5RC1
+     */
+    @Unstable
+    default long getSingleJobThreadKeepAliveTime()
+    {
+        return 60000L;
+    }
+
+    /**
+     * @return the duration in milliseconds for the thread keep alive in our grouped job thread executor.
+     * @see java.util.concurrent.ThreadPoolExecutor#setKeepAliveTime(long, TimeUnit).
+     * @since 12.5RC1
+     */
+    @Unstable
+    default long getGroupedJobThreadKeepAliveTime()
+    {
+        return 60000L;
+    }
 }
--- a/xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/DefaultGroupedJobInitializerManager.java
+++ b/xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/DefaultGroupedJobInitializerManager.java
@@ -19,6 +19,7 @@
  */
 package org.xwiki.job.internal;
 
+import java.util.Collections;
 import java.util.List;
 
 import javax.inject.Inject;
@@ -33,6 +34,7 @@ import org.xwiki.cache.CacheException;
 import org.xwiki.cache.CacheManager;
 import org.xwiki.cache.config.LRUCacheConfiguration;
 import org.xwiki.component.annotation.Component;
+import org.xwiki.component.event.ComponentDescriptorRemovedEvent;
 import org.xwiki.component.manager.ComponentLookupException;
 import org.xwiki.component.manager.ComponentManager;
 import org.xwiki.component.phase.Initializable;
@@ -41,6 +43,8 @@ import org.xwiki.job.GroupedJobInitializer;
 import org.xwiki.job.GroupedJobInitializerManager;
 import org.xwiki.job.JobGroupPath;
 import org.xwiki.job.JobManagerConfiguration;
+import org.xwiki.observation.EventListener;
+import org.xwiki.observation.event.Event;
 
 /**
  * A component dedicated to find the appropriate {@link GroupedJobInitializer} based on a {@link JobGroupPath}.
@@ -50,7 +54,7 @@ import org.xwiki.job.JobManagerConfiguration;
  */
 @Component
 @Singleton
-public class DefaultGroupedJobInitializerManager implements GroupedJobInitializerManager, Initializable
+public class DefaultGroupedJobInitializerManager implements GroupedJobInitializerManager, Initializable, EventListener
 {
     @Inject
     @Named("context")
@@ -138,5 +142,28 @@ public class DefaultGroupedJobInitializerManager implements GroupedJobInitialize
         }
 
         return result;
+    }
+
+    @Override
+    public String getName()
+    {
+        return getClass().getSimpleName();
+    }
+
+    @Override
+    public List<Event> getEvents()
+    {
+        return Collections.singletonList(new ComponentDescriptorRemovedEvent());
+    }
+
+    @Override
+    public void onEvent(Event event, Object source, Object data)
+    {
+        ComponentDescriptorRemovedEvent componentDescriptorRemovedEvent = (ComponentDescriptorRemovedEvent) event;
+
+        // Flush the cache in case a GroupedJobInitializer component has been removed
+        if (componentDescriptorRemovedEvent.getRoleType() == GroupedJobInitializer.class) {
+            this.cachedInitializers.removeAll();
+        }
     }
 }
--- a/xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/DefaultGroupedJobInitializerManager.java
+++ b/xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/DefaultGroupedJobInitializerManager.java
@@ -1,0 +1,142 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.job.internal;
+
+import java.util.List;
+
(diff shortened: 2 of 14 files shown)


surli:
Personal reminder: I also need to edit `xwiki.properties.vm` in platform with the new properties once it's merged.

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/JobGroupPathLockTree.java
```diff
@@ -32,31 +38,104 @@
...
 public class JobGroupPathLockTree
 {
-    private final Map<JobGroupPath, ReadWriteLock> tree = new ConcurrentHashMap<>();
+    @Inject
+    private GroupedJobInitializerManager groupedJobInitializerManager;
+
+    /**
+     * A specific concurrency implementation for managing Semaphore with Read/Write lock capabilities.
+     */
+    private class ReadWriteSemaphore
```
tmortagne, line 51: Might worth putting that in its own file ideally somewhere in commons maybe since it's very generic.

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/JobGroupPathLockTree.java
```diff
@@ -32,31 +38,104 @@
...
+            this.semaphore.acquireUninterruptibly(this.readCounter + 1);
+        }
+
+        void unlockWrite()
+        {
+            this.writeCounterLock.lock();
+            this.writeCounter--;
+
+            if (this.writeCounter == 0) {
+                this.writeCounterLock.unlock();
```
tmortagne, line 80: You have the exact same line in both branches of the if so I guess it could be before it.

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/JobGroupPathLockTree.java
surli: Might make sense. Not as a public API though but internal, right?

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/JobManagerConfiguration.java
```diff
@@ -42,4 +44,36 @@
...
     int getJobStatusCacheSize();
+
+    /**
+     * @return the number of {@link GroupedJobInitializer} to keep in cache.
+     * @since 12.5RC1
+     */
+    @Unstable
+    default int getGroupedJobInitializerCacheSize()
+    {
+        return 10;
```
surli, line 55: Do you agree with this default value for the cache of groupedJobInitializer @tmortagne ?

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/JobGroupPathLockTree.java
tmortagne: Why not public ? So that we can ditch it if we find a lib that does this ?

surli: > So that we can ditch it if we find a lib that does this ?

yep.

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/JobManagerConfiguration.java
tmortagne: It's probably a bit low. I doubt GroupedJobInitializer use much memory so we could easily increase it a bit by default (like 100 maybe). Would need to check how many nodes are created after a few jobs are run in XWiki Standard. On extension side there is the root "extension" node and then 1 per namespace (so in practice 1 par wiki). It's potentially a lot more for the refactoring module since you have the root "refactoring" node and then one node for each element of each EntityReference which is manipulated.

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/DefaultGroupedJobInitializerManager.java
```diff
@@ -0,0 +1,142 @@
...
+    @Inject
+    private GroupedJobInitializer defaultGoupedJobInitializer;
+
+    @Inject
+    private CacheManager cacheManager;
+
+    @Inject
+    private JobManagerConfiguration configuration;
+
+    private Cache<GroupedJobInitializer> cachedInitializers;
```
tmortagne, line 71: Should maybe flush that cache when a GroupedJobInitializer component is registered/unregistered (ComponentDescriptorAddedEvent/ComponentDescriptorRemovedEvent).

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/JobManagerConfiguration.java
surli: > It's probably a bit low. I doubt GroupedJobInitializer use much memory so we could easily increase it a bit by default (like 100 maybe).

OK let's do that: my first move was to put 1000 but it seemed a bit high back then :) Especially since I don't expect we'll have many GroupedJobInitializer, but maybe I'm wrong there.

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/JobGroupPathLockTree.java
tmortagne: OK then good point.

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/JobGroupPathLockTree.java
```diff
@@ -32,31 +38,104 @@
...
+                this.semaphore.release(this.readCounter + 1);
+            } else {
+                this.writeCounterLock.unlock();
+                this.semaphore.release();
+            }
+        }
+
+        void lockRead()
+        {
+            this.readCounter++;
```
surli, line 90: @tmortagne do you think I should protect them too with a dedicated lock, like I did for `writeCounter`? I started to do it, but then removed it after I got some trouble (deadlock) with them. I didn't put them back because I thought they were useless since we don't use that value for testing a condition, but we definitely use it for taking the semaphore permits, so now I'm hesitating about it...

surli: I guess I'd need to protect it, volatile won't protect them for "read-update-write" and so I could have a race condition with a lockRead/unlockRead leading to a wrong counter value.

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/DefaultGroupedJobInitializerManager.java
surli: I hesitated about the way to deal with that cache: right now, if you check on #search method I check first on the cache and then build back it when I iterate over the components. My first move was to build the cache once and to rebuilt it in case of component added/removed but not sure how to ensure that the cache is big enough: what happens if I only rely on the cache and there's 101+ components? Some of them will never be cached and so I won't be able to reach them. Right now the strategy is more lazy: I put info in the cache as I need them and I never rely entirely on it, so I don't think there's a real need for listening on events.

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/JobGroupPathLockTree.java
surli: FYI I changed it to protect it.

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/DefaultGroupedJobInitializerManager.java
tmortagne: Lazy is the way to go but the problem is that you might rely on a GroupedJobInitializer instance which does not work anymore because it was unloaded for example.

surli: Right, so I need to flush the cache in case of removed event only then.

tmortagne: I would still clean on both ComponentDescriptorAddedEvent and ComponentDescriptorRemovedEvent.

surli: Can do but I don't really see the point for cleaning on added event: we'd lose some values that are still correct in the cache...

tmortagne: You can imagine installing a new GroupedJobInitializer being associated with a path which is already in the cache because it was inherited from some parent because there was no GroupedJobInitializer back then for this specific path.

surli: > a new GroupedJobInitializer being associated with a path which is already in the cache because it was inherited from some parent because there was no GroupedJobInitializer back then for this specific path.

I only keep the exact JobGroupPath mapping in cache (the value given by the GroupedJobInitializer#getId), so that scenario is not possible.

surli: So I did it only for remove event, it's easy to improve later if we want. See: https://github.com/xwiki/xwiki-commons/pull/96/files#diff-7024b53c2653ba2d321478a35653d515R156

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/DefaultGroupedJobInitializerManager.java
```diff
@@ -0,0 +1,142 @@
...
+     *      matches the path. If it matches stop iterating.
+     * @param path the searched path.
+     * @return an initializer matching the path or null if no initializer has been found.
+     */
+    private GroupedJobInitializer search(JobGroupPath path)
+    {
+        GroupedJobInitializer result = this.cachedInitializers.get(path.toString());
+        if (result == null) {
+            try {
+                List<GroupedJobInitializer> initializers =
```
tmortagne, line 104: Wouldn't it make more sense from performance point of view to give this list to `search()` instead of recreating it for each element of the path ? The list is always the same during the `for` in getGroupedJobInitializer.

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/DefaultGroupedJobInitializerManager.java
tmortagne: > I only keep the exact JobGroupPath mapping in cache (the value given by the GroupedJobInitializer#getId), so that scenario is not possible.

I was expecting inherited values to be put in the cache too to to make search faster.

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/DefaultGroupedJobInitializerManager.java
surli: Might make sense since we don't trust much the cache and we are always looping on the components to find an exact match which might be the exception: more likely we'll fallback on a parent initializer so we'll loop there at least once but maybe twice or more.

xwiki-commons-core/xwiki-commons-job/src/main/java/org/xwiki/job/internal/DefaultGroupedJobInitializerManager.java
surli: > I was expecting inherited values to be put in the cache too to make search faster

The rationale was actually that you won't call twice the initializer with the same jobGroupPath so it felt useless, but I'm realizing while writing this that it's actually the manager and not the initializer here and you'll definitely call it at least twice to get the initializer (once for executor and once for the JobGroupPathLockTree), so my argument is not really good anymore...

PR #100, opened 2020-08-23
[misc] exclude org.reflections dependency from checkstyle 

romani:
to not conflict with checkstyle's version

library will be introduced at https://github.com/checkstyle/checkstyle/pull/8743


tmortagne:
Thank for the contibution @romani. Note that there is a serious regression in Reflections 0.9.12 (https://github.com/ronmamo/reflections/issues/273) and it seems to be pretty much dead so Checkstyle might want to reconsider using it.

PR #101, opened 2020-09-01
XCOMMONS-2016: Add a new escape method that only escape the minimal amount of XML special characters

manuelleduc:
https://jira.xwiki.org/browse/XCOMMONS-2016

### DIFF
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
@@ -51,6 +51,7 @@ import org.w3c.dom.ls.LSSerializer;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
 import org.xml.sax.XMLReader;
+import org.xwiki.stability.Unstable;
 
 /**
  * XML Utility methods.
@@ -266,6 +267,41 @@ public final class XMLUtils
                     break;
                 case '{':
                     result.append(LCURL);
+                    break;
+                default:
+                    result.append(c);
+            }
+        }
+        return result.toString();
+    }
+
+    /**
+     *  Escapes some XML special characters (&amp; and &lt;) in a <code>String</code> using numerical XML
+     *  entities, so that the resulting string can safely be used in HTML pages.
+     *
+     * @param content the text to escape, may be {@code null}
+     * @return a new escaped {@code String}, {@code null} if {@code null} input
+     * @since 12.8RC1
+     */
+    @Unstable
+    public static String minimalEscape(Object content)
+    {
+        if (content == null) {
+            return null;
+        }
+        String str = String.valueOf(content);
+        // Initializes a string builder with an initial capacity 1.1 times greater than the initial content to account
+        // for special character substitutions.
+        int contentLength = str.length();
+        StringBuilder result = new StringBuilder((int) (contentLength * 1.1));
+        for (int i = 0; i < contentLength; ++i) {
+            char c = str.charAt(i);
+            switch (c) {
+                case '&':
+                    result.append(AMP);
+                    break;
+                case '<':
+                    result.append(LT);
                     break;
                 default:
                     result.append(c);
--- a/xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/XMLUtilsTest.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/XMLUtilsTest.java
@@ -66,6 +66,19 @@ public class XMLUtilsTest
     }
 
     @Test
+    void escapeMinimal()
+    {
+        String actual = XMLUtils.minimalEscape("a < a' && a' < a\" => a < a\"");
+        assertEquals("a &#60; a' &#38;&#38; a' &#60; a\" => a &#60; a\"", actual);
+    }
+
+    @Test
+    void escapeMinimalNull()
+    {
+        assertNull(XMLUtils.minimalEscape(null));
+    }
+
+    @Test
     void escapeApos()
     {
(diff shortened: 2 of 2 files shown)


xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -274,6 +274,39 @@ public static String escapeAttributeValue(Object content)
...
 
+    /**
+     *  Escapes some XML special characters (&amp; and &lt;) in a <code>String</code> using numerical XML
+     *  entities, so that the resulting string can safely be used in HTML pages.
+     *
+     * @param content the text to escape, may be {@code null}
+     * @return a new escaped {@code String}, {@code null} if {@code null} input
+     * @since 12.8RC1
+     */
+    public static String minimalEscape(Object content)
```
tmortagne, line 285: Missing Unstable attribute.

tmortagne (approved):

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
manuelleduc: Fixed

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -274,6 +275,40 @@ public static String escapeAttributeValue(Object content)
...
+     * @since 12.8RC1
+     */
+    @Unstable
+    public static String minimalEscape(Object content)
+    {
+        if (content == null) {
+            return null;
+        }
+        String str = String.valueOf(content);
+        StringBuilder result = new StringBuilder((int) (str.length() * 1.1));
```
mflorea, line 293: Any magic number deserves a comment.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -274,6 +275,40 @@ public static String escapeAttributeValue(Object content)
...
+        int length = str.length();
```
mflorea, line 294: You're using ``str.length()`` above also so I would declare the variable before.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/XMLUtilsTest.java
```diff
@@ -65,6 +65,24 @@ void escape()
...
+    @Test
+    void escapeMinimal()
+    {
+        String escapedText = XMLUtils.minimalEscape("a < a' && a' < a\" => a < a\"");
+
+        assertFalse(escapedText.contains("<"), "Failed to escape <");
+        assertFalse(escapedText.contains("&&"), "Failed to escape &");
+        assertTrue(escapedText.contains(">"), "Failed to keep >");
+        assertTrue(escapedText.contains("'"), "Failed to keep '");
+        assertTrue(escapedText.contains("\""), "Failed to keep \"");
```
mflorea, line 77: I find it safer to assert the exact output and not whether the output contains some strings. It's not like the input string is that long. It's also easier to see the difference between the expected and the actual value when the test fails.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
manuelleduc: Documented here https://github.com/xwiki/xwiki-commons/pull/101/files#diff-c1eed56adbad8fd67a8e196907ce032fR293

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
manuelleduc: Factored here https://github.com/xwiki/xwiki-commons/pull/101/files#diff-c1eed56adbad8fd67a8e196907ce032fR295

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/XMLUtilsTest.java
manuelleduc: I reused the testing style of the test class, but I agree with you. Test updated here https://github.com/xwiki/xwiki-commons/pull/101/files#diff-4d325ecb5d796d402110ecb9c1edfdf7R72

mflorea (approved):

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -274,6 +275,41 @@ public static String escapeAttributeValue(Object content)
...
+                    result.append(AMP);
+                    break;
+                case '<':
+                    result.append(LT);
+                    break;
+                default:
+                    result.append(c);
+            }
+        }
+        return result.toString();
```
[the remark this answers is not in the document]
manuelleduc: I agree this is not the usual way to escape html. But in some context (see https://jira.xwiki.org/browse/XWIKI-9680), if we fully escape the string the content is badly displayed for the end user (IIUC, it is caused by a miss use  of `PropertyClass#displayView` method).
This method is an attempt to find a trade off between the needed string escaping on one side, and the user experience on another side.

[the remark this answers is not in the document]
vmassol: Then I'm not sure why we're making this public if it's not the right way it should be...

PR #102, opened 2020-09-07
XCOMMONS-2016: Add a new escape method that only escape the minimal amount of XML special characters

manuelleduc:
https://jira.xwiki.org/browse/XCOMMONS-2016

### DIFF
--- a/xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/EscapeTool.java
+++ b/xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/EscapeTool.java
@@ -23,6 +23,7 @@ import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.util.Collection;
 import java.util.Map;
+import java.util.Objects;
 
 import org.apache.commons.codec.EncoderException;
 import org.apache.commons.codec.net.BCodec;
@@ -68,7 +69,7 @@ public class EscapeTool extends org.apache.velocity.tools.generic.EscapeTool
     @Override
     public String xml(Object content)
     {
-        return XMLUtils.escape(content);
+        return XMLUtils.escape(Objects.toString(content, null));
     }
 
     /**
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
@@ -225,6 +225,7 @@ public final class XMLUtils
      * @param content the text to escape, may be {@code null}
      * @return a new escaped {@code String}, {@code null} if {@code null} input
      */
+    @Deprecated
     public static String escape(Object content)
     {
         return escapeAttributeValue(content);
@@ -276,15 +277,15 @@ public final class XMLUtils
     }
 
     /**
-     *  Escapes some XML special characters (&amp; and &lt;) in a <code>String</code> using numerical XML
-     *  entities, so that the resulting string can safely be used in HTML pages.
+     * Escapes XML special characters in a {@code String} using numerical XML entities, so that the resulting string
+     * can safely be used as an XML text value. Specifically, escapes &lt;, and &amp;.
      *
      * @param content the text to escape, may be {@code null}
      * @return a new escaped {@code String}, {@code null} if {@code null} input
      * @since 12.8RC1
      */
     @Unstable
-    public static String minimalEscape(Object content)
+    public static String escapeElementText(Object content)
     {
         if (content == null) {
             return null;
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
@@ -22,6 +22,7 @@ package org.xwiki.xml;
 import java.io.IOException;
 import java.io.StringReader;
 import java.io.StringWriter;
+import java.util.Objects;
 import java.util.regex.Pattern;
 
 import javax.xml.parsers.DocumentBuilderFactory;
@@ -246,7 +247,7 @@ public final class XMLUtils
     @Deprecated
     public static String escape(Object content)
     {
-        return escapeAttributeValue(content);
+        return escapeAttributeValue(Objects.toString(content, null));
     }
 
     /**
@@ -272,7 +273,9 @@ public final class XMLUtils
      * @return a new escaped {@code String}, {@code null} if {@code null} input
      * @see #escapeAttributeValue(String)
      * @see #escapeElementText(String)
+     * @since 12.8RC1
      */
+    @Unstable
     public static String escape(String content)
     {
         return escapeAttributeValue(content);
@@ -318,7 +321,9 @@ public final class XMLUtils
      *
      * @param content the text to escape, may be {@code null}
      * @return a new escaped {@code String}, {@code null} if {@code null} input
+     * @since 12.8RC1
      */
+    @Unstable
     public static String escapeAttributeValue(String content)
     {
         if (content == null) {
@@ -416,7 +421,9 @@ public final class XMLUtils
      *
      * @param content the text to escape, may be {@code null}
      * @return a new escaped {@code String}, {@code null} if {@code null} input
+     * @since 12.8RC1
      */
+    @Unstable
     public static String escapeElementContent(String content)
     {
         if (content == null) {
@@ -469,7 +476,9 @@ public final class XMLUtils
      *
      * @param content the text to decode, may be {@code null}
      * @return unescaped content, {@code null} if {@code null} input
+     * @since 12.8RC1
      */
+    @Unstable
     public static String unescape(String content)
     {
         if (content == null) {
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
@@ -218,20 +218,27 @@ public final class XMLUtils
     }
 
     /**
+     * Escapes all the XML special characters and a XWiki Syntax 2.0+ special character (i.e., <code>{</code>) in a
+     * {@code String} using numerical XML entities.
+     * For instance, {@code <b>{{html}}$x{{/html}}</b>} will be escaped and can thus be put inside an
+     * XML tag or an XML attribute.
+     * For instance, in a div tag
+     * <code>&lt;div&gt;&amp;#60;b&amp;#62;&amp;#123;&amp;#123;html}}$x&amp;#123;&amp;#123;/html}}&amp;#60;/b&amp;#62;
+     *&lt;/div&gt;</code>
+     * or in the value attribute of an input tag
+     * <code>&lt;input
+     * value=&quot;&amp;#60;b&amp;#62;&amp;#123;&amp;#123;html}}$x&amp;#123;&amp;#123;/html}}&amp;#60;/b&amp;#62;&quot;
+     * /&gt;</code>.
+     * Specifically, escapes &lt;, &gt;, ", ', &amp; and {.
+     * Left curly bracket is included here to protect against {{/html}} in xwiki 2.x syntax.
      * <p>
-     * Escapes all the XML special characters and a xwiki 2.x special character (i.e., <code>{</code>) in a
-     * {@code String} using numerical XML entities. Specifically, escapes &lt;, &gt;, ", ', &amp; and {. Left curly
-     * bracket is included here to protect against {{/html}} in xwiki 2.x syntax.
-     * </p>
-     * <p>
-     * Note that is is preferable to use {@link XMLUtils#escapeAttributeValue(Object)} when the content is used as 
-     * an XML tag attribute, and {@link XMLUtils#escapeElementText(Object)} when the content is used as an XML text.
-     * </p>
+     * Note that is is preferable to use {@link #escapeAttributeValue(Object)} when the content is used as
+     * an XML tag attribute, and {@link #escapeElementText(String)} when the content is used as an XML text.
      *
      * @param content the text to escape, may be {@code null}
      * @return a new escaped {@code String}, {@code null} if {@code null} input
-     * @see XMLUtils#escapeAttributeValue(Object)
-     * @see XMLUtils#escapeElementText(Object) 
+     * @see #escapeAttributeValue(Object)
+     * @see #escapeElementText(String)
      */
     public static String escape(Object content)
     {
@@ -285,28 +292,27 @@ public final class XMLUtils
 
     /**
      * Escapes XML special characters in a {@code String} using numerical XML entities, so that the resulting string
-     * can safely be used as an XML element text value. For instance, {@code Jim & John} will be escaped and can thus
-     * be put inside an XML tag, such as the {@code p} tag, as in {@code <p>Jim &amp; John</p>}. Specifically, escapes
-     * &lt; to {@code &lt;}, and &amp; to {@code &amp;}.
+     * can safely be used as an XML element text value.
+     * For instance, {@code Jim & John} will be escaped and can thus be put inside an XML tag, such as the {@code p}
+     * tag, as in {@code <p>Jim &amp; John</p>}.
+     * Specifically, escapes &lt; to {@code &lt;}, and &amp; to {@code &amp;}.
      *
-     * @param content the text to escape, may be {@code null}. The content is converted to a {@link String} using
-     * {@link String#valueOf(Object)} before escaping.
+     * @param content the text to escape, may be {@code null}.
      * @return a new escaped {@code String}, {@code null} if {@code null} input
      * @since 12.8RC1
      */
     @Unstable
-    public static String escapeElementText(Object content)
+    public static String escapeElementText(String content)
     {
         if (content == null) {
             return null;
         }
-        String str = String.valueOf(content);
         // Initializes a string builder with an initial capacity 1.1 times greater than the initial content to account
         // for special character substitutions.
-        int contentLength = str.length();
+        int contentLength = content.length();
         StringBuilder result = new StringBuilder((int) (contentLength * 1.1));
         for (int i = 0; i < contentLength; ++i) {
-            char c = str.charAt(i);
+            char c = content.charAt(i);
             switch (c) {
                 case '&':
                     result.append(AMP);
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
@@ -218,14 +218,21 @@ public final class XMLUtils
     }
 
     /**
-     * Escapes all the XML special characters in a <code>String</code> using numerical XML entities. Specifically,
-     * escapes &lt;, &gt;, ", ', &amp; and {.  Left curly bracket is included here to protect against {{/html}} in
-     * xwiki 2.x syntax.
+     * <p>
+     * Escapes all the XML special characters and a xwiki 2.x special character (i.e., <code>{</code>) in a
+     * {@code String} using numerical XML entities. Specifically, escapes &lt;, &gt;, ", ', &amp; and {. Left curly
+     * bracket is included here to protect against {{/html}} in xwiki 2.x syntax.
+     * </p>
+     * <p>
+     * Note that is is preferable to use {@link XMLUtils#escapeAttributeValue(Object)} when the content is used as 
+     * an XML tag attribute, and {@link XMLUtils#escapeElementText(Object)} when the content is used as an XML text.
+     * </p>
      *
      * @param content the text to escape, may be {@code null}
      * @return a new escaped {@code String}, {@code null} if {@code null} input
+     * @see XMLUtils#escapeAttributeValue(Object)
+     * @see XMLUtils#escapeElementText(Object) 
      */
-    @Deprecated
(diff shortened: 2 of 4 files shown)


xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -225,6 +225,7 @@ public static String unescapeXMLComment(String content)
      * @param content the text to escape, may be {@code null}
      * @return a new escaped {@code String}, {@code null} if {@code null} input
      */
+    @Deprecated
```
vmassol, line 228: As I mentioned on some chat, I'm not sure we wish to deprecate this. Note that whenever a deprecation is done it also requires a `@deprecated` javadoc tag with explanation as to what should be used and in which version it's been deprecated.

tmortagne: There is no reason to deprecate this.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -276,15 +277,15 @@ public static String escapeAttributeValue(Object content)
     }
 
     /**
-     *  Escapes some XML special characters (&amp; and &lt;) in a <code>String</code> using numerical XML
-     *  entities, so that the resulting string can safely be used in HTML pages.
+     * Escapes XML special characters in a {@code String} using numerical XML entities, so that the resulting string
+     * can safely be used as an XML text value. Specifically, escapes &lt;, and &amp;.
```
vmassol, line 281: "as an XML text value." --> "as an XML element text value (e.g. {@code Jim & John} will be escaped and can thus be put inside an XML tag, such as the {@code p} tag, as in {@code <p>Jim &amp; John</p>}).".

Should use `{@code ...}` constructs for `&lt;` and `&amp;`.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -276,15 +277,15 @@ public static String escapeAttributeValue(Object content)
...
      *
      * @param content the text to escape, may be {@code null}
      * @return a new escaped {@code String}, {@code null} if {@code null} input
      * @since 12.8RC1
      */
     @Unstable
-    public static String minimalEscape(Object content)
+    public static String escapeElementText(Object content)
```
vmassol, line 288: I think the javadoc for the content param should also be improved and we should explain why we accept an Object and not a String and what we do to it (say that we convert it to a String).

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
```diff
@@ -289,4 +291,18 @@ public static void stripFirstElementInside(Document document, String parentTagNa
...
+    /**
+     * Escapes HTML special characters in a {@code String} using numerical HTML entities, so that the resulting string
+     * can safely be used as an HTML content. Specifically, escapes &lt;, and &amp;.
+     *
+     * @param content the text to escape, may be {@code null}
+     * @return a new escaped {@code String}, {@code null} if {@code null} input
+     * @since 12.8RC1
+     */
+    @Unstable
+    public static String escapeElementContent(Object content)
```
vmassol, line 304: Is it called Elements in HTML? We need to check this. Could be Tags (https://www.w3schools.com/tags/tag_div.asp calls it tags at least).

tmortagne: It's called an "element" in all the XML APIs I can think of at least.

vmassol: > It's called an "element" in all the XML APIs I can think of at least.

Yes but we're in HTMLUtils here. And it has its own spec and terminology.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
vmassol: Well there's a reason :) It's very unprecise/misleading and leads to extra escapes that are not needed in some situations. I'm fine to keep it but the javadoc needs to be improved to explain that (the extra escapes) and we need to recommended using the other 2 methods whenever possible.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
manuelleduc: I read some more doc and `Content` is too broad as it applied both to children tags and texts. `escapeElementText` is probably more accurate (and aligns with the XML vocabulary).

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
manuelleduc: Documentation of the content parameter updated here https://github.com/xwiki/xwiki-commons/pull/102/files#diff-c1eed56adbad8fd67a8e196907ce032fR292

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
manuelleduc: Deprecated annotation removed and javadoc updated here https://github.com/xwiki/xwiki-commons/pull/102/files#diff-c1eed56adbad8fd67a8e196907ce032fR221
- Some more explanations of why { is escaped
- Explanations of why is is best to use `escapeAttributeValue` and `escapeElementText` instead of `escape`

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
manuelleduc: Javadoc updated here https://github.com/xwiki/xwiki-commons/pull/102/files#diff-c1eed56adbad8fd67a8e196907ce032fR287

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
manuelleduc: Method renamed to `escapeElementText` and javadoc updated here https://github.com/xwiki/xwiki-commons/pull/102/files#diff-6866bfd5f73fabc8380825bf60260423R307

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -218,12 +218,20 @@ public static String unescapeXMLComment(String content)
     }
 
     /**
-     * Escapes all the XML special characters in a <code>String</code> using numerical XML entities. Specifically,
-     * escapes &lt;, &gt;, ", ', &amp; and {.  Left curly bracket is included here to protect against {{/html}} in
-     * xwiki 2.x syntax.
+     * <p>
+     * Escapes all the XML special characters and a xwiki 2.x special character (i.e., <code>{</code>) in a
```
vmassol, line 222: What is a "xwiki 2.x special character"? Maybe you mean "a XWiki Syntax 2.0+ special character"? Now this is already mentioned at the end of the sentence. 

Note: I think `</p>` leads to error in the javadoc generation but I'm not sure. I don't think tags must be closed in javadoc (it's not pure HTML). See for ex https://blog.joda.org/2012/11/javadoc-coding-standards.html

I think the javadoc still missing explanation as to why you would use this method. See my proposals on the chat.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
```diff
@@ -289,4 +291,21 @@ public static void stripFirstElementInside(Document document, String parentTagNa
...
         }
     }
+
+    /**
+     * Escapes HTML special characters in a {@code String} using numerical HTML entities, so that the resulting string
+     * can safely be used as an HTML content text value. For instance, {@code Jim & John} will be escaped and can thus
+     * be put inside a HTML tag, such as the {@code p} tag, as in {@code <p>Jim &amp; John</p>}. Specifically, escapes
+     * &lt; to {@code &lt;}, and &amp; to {@code &amp;}.
+     *
+     * @param content the text to escape, may be {@code null}. The content is converted to a {@link String} using
```
vmassol, line 301: Still missing the why? Why did you add a signature that takes an Object instead of a String!

manuelleduc: I made `content` an `Object` to stay consistent with the other escape methods. But AFAIK we would be fine with a `content` of type `String`. Would it be ok to change the signature of `escapeElementText` to `public static String escapeElementText(String content)`?

vmassol: > I made content an Object to stay consistent with the other escape methods.

But that's not a good-enough reason :) Why did they do that? Imagine they made a mistake, you'd just be continuing the mistake.

vmassol: I think it comes from Velocity.

See:
* Taken from https://github.com/xwiki/xwiki-platform/commit/ed3a5467b08c7069959076ace8e40a2c17bb253c#diff-83c0483287bd455e80eb573959efc65bR47
* And move to https://github.com/xwiki/xwiki-platform/commit/ed3a5467b08c7069959076ace8e40a2c17bb253c#diff-2c6a53f98aefb9159a01c42a82e54255R177

Now need to understand why Velocity tool is using Object. And whether it's needed for a Java API (I don't think so).

vmassol: Not sure what to do. I think i'd be in favor of using String for the new method instead of Object. WDYT?

manuelleduc: - We control the places where the method is called (and they are all passing string values)
- IMHO, it is a good practice to force the caller to sanitize its input before (instead of dealing with unexpected value in a method).

I'm in favor of using String for the new method.

vmassol: Yep, let's see if @tmortagne and @surli are ok on this too.

vmassol: Otherwise it would mean that everywhere we take a String in APIs we should move to Object and convert to a String. Feel not great and not typed ;)

vmassol: We should probably introduce new signatures for existing methods taking an Object and deprecate existing ones (AFAIK specific types are favored over Object in Java so it should work).

tmortagne: I can understand it for Velocity but I don't like this Object parameter for a Java utility method.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
manuelleduc: Javadoc update to use better writing standard + adding more explanation on why each method should be used for

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
manuelleduc: Each method taking an Object in parameter is now deprecated and is overridden with a method taking a String in parameter

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -349,24 +448,42 @@ public static String escapeElementContent(Object content)
...
      * ones that affect the resulting markup.
      *
-     * @param content the text to decode, may be {@code null}
+     * @param content the text to decode, may be {@code null}. The content is converted to {@code String} using
+     * {@link String#valueOf(Object)} before escaping.
      * @return unescaped content, {@code null} if {@code null} input
+     * @deprecated since 12.8RC1, use {@link #unescape(String)} instead
      */
+    @Deprecated
     public static String unescape(Object content)
```
vmassol, line 457: I think it should be easy in your IDE to find all existing calls to `unescape(Object)` in our code base and replace them with calls to `unescape(String)`. This would allow you/us to move the deprecated method to legacy (see "Second Step" in https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HBackwardCompatibility).

WDYT?

manuelleduc: Doing it. 
Also:
1. I forgot to add javadoc `@since` for the new methods (fixing)
2. Should I add an `@Unstable` annotation on the new methods?

vmassol: > Should I add an @Unstable annotation on the new methods?

Yes

vmassol: > Doing it.

Cool, thanks!

manuelleduc: Deprecated and since added.

manuelleduc: Call to the deprecated methods updated in https://github.com/xwiki/xwiki-platform/pull/1384

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/EscapeTool.java
```diff
@@ -68,7 +69,7 @@
     @Override
     public String xml(Object content)
     {
-        return XMLUtils.escape(content);
+        return XMLUtils.escape(Objects.toString(content, null));
```
surli, line 72: Second argument is completely useless there no? According to javadoc it's exactly what does `Objects.toString` with only one argument.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -218,37 +219,121 @@ public static String unescapeXMLComment(String content)
...
+     * /&gt;</code>.
+     * <p>
+     * Specifically, escapes &lt;, &gt;, ", ', &amp; and {.
+     * <p>
+     * Note that is is preferable to use {@link #escapeAttributeValue(String)} when the content is used as
+     * an XML tag attribute, and {@link #escapeElementText(String)} when the content is used as an XML text.
      *
-     * @param content the text to escape, may be {@code null}
+     * @param content the text to escape, may be {@code null}. The content is converted to {@code String} using
+     * {@link String#valueOf(Object)} before escaping.
```
surli, line 241: You're actually using `Objects.toString()` here.

surli (approved):

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -314,20 +401,39 @@ public static String minimalEscape(Object content)
...
+    /**
+     * Escapes the XML special characters in a <code>String</code> using numerical XML entities, so that the resulting
+     * string can safely be used as an XML text node. Specifically, escapes &lt;, &gt;, and &amp;.
+     *
+     * @param content the text to escape, may be {@code null}
+     * @return a new escaped {@code String}, {@code null} if {@code null} input
+     * @since 12.8RC1
+     */
+    @Unstable
+    public static String escapeElementContent(String content)
```
mflorea, line 427: It's not clear at all from the JavaDoc **why** we need both ``escapeElementText`` and ``escapeElementContent``. The JavaDoc explains what each method does and I understand that the only difference is that ``escapeElementText`` doesn't escape ``>`` (``&gt;``) but when would you use one versus the other? When is it OK to not escape ``>`` and when is it mandatory to escape it?

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/EscapeTool.java
manuelleduc:
```
jshell> Objects.toString(null);
$1 ==> "null"

jshell> Objects.toString(null, null);
$2 ==> null
```

`Objects#toString(Object)` returns the `"null"` string in case of a null parameter.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
tmortagne: Actually escaping `>` is never really needed so I don't understand why we need escapeElementContent at all.

manuelleduc: Would it be ok to deprecate `escapeElementContent` and suggest to use `escapeElementText` instead?

tmortagne: Sounds good.

manuelleduc: Hum, in this case we are in a odd situation when `public static String escapeElementContent(String content)` will be both `@Unstable` and `@Deprecated`. Does deprecated takes precedence on unstable?

tmortagne: I don't understand what you mean, `escapeElementContent(String content)` should not exist at all.

manuelleduc: `escapeElementContent` deprecated and `escapeElementText` recommended instead

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -218,37 +219,121 @@ public static String unescapeXMLComment(String content)
...
      * @return a new escaped {@code String}, {@code null} if {@code null} input
+     * @see #escapeAttributeValue(String)
+     * @see #escapeElementText(String)
+     * @deprecated since 12.8RC1, use {@link #escape(String)} instead
      */
+    @Deprecated
     public static String escape(Object content)
     {
-        return escapeAttributeValue(content);
+        return escapeAttributeValue(Objects.toString(content, null));
```
mflorea, line 250: I would call ``escape`` here to be in line with the ``@deprecated`` message and to make sure that both ``escape`` methods behave the same (someone may update ``escape(String)`` without updating the deprecated ``escape(Object)``). This is minor but it's a best practice IMO.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -349,24 +439,44 @@ public static String escapeElementContent(Object content)
...
-        str = QUOT_PATTERN.matcher(str).replaceAll("\"");
-        str = LT_PATTERN.matcher(str).replaceAll("<");
-        str = GT_PATTERN.matcher(str).replaceAll(">");
-        str = AMP_PATTERN.matcher(str).replaceAll("&");
-        str = LCURL_PATTERN.matcher(str).replaceAll("{");
+        return unescape(String.valueOf(content));
+    }
+
+    /**
+     * Unescape encoded special XML characters. Only &gt;, &lt; &amp;, ", ' and { are unescaped, since they are the only
```
tmortagne, line 458: There is no reason to provide a custom unecaper, anyone with the need to unecape XML should use one of the many existing XML unescaper.

manuelleduc: Unescaper deprecated and `org.apache.commons.text.StringEscapeUtils#unescapeXml(String)` suggested insted.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
manuelleduc: I agree, `escape(Object)` is now calling `escape(String)` which itself calls `escapeAttributeValue(String)`.

tmortagne (approved):

mflorea (approved):

PR #104, opened 2020-09-18
XCOMMONS-2025: use a new EntityResolver based on css4j's DefaultEntityResolver

carlosame:
Requires css4j 3.0 or higher. It preserves XWiki's ability to connect to remote DTD resources (which some users may depend on), doing it more securely (making sure that they are really DTDs).

To actually use this new resolver:

```diff
@@ -5,7 +5,7 @@ org.xwiki.xml.internal.html.filter.LinkFilter
 org.xwiki.xml.internal.html.filter.BodyFilter
 org.xwiki.xml.internal.html.filter.ControlCharactersFilter
 org.xwiki.xml.internal.html.filter.AttributeFilter
 org.xwiki.xml.internal.html.filter.UniqueIdFilter
 org.xwiki.xml.internal.html.DefaultHTMLCleaner
-org.xwiki.xml.internal.LocalEntityResolver
+org.xwiki.xml.internal.DefaultEntityResolver
 org.xwiki.xml.internal.DefaultXMLReaderFactory
\ No newline at end of file
```


vmassol:
@carlosame thanks a lot for all these PRs! We're just finishing the 12.8 release (releasing RC1 Monday and then final) and then we can apply your PRs for master (which will be 12.9). So expect a small delay of up to one week more before we apply them. Just wanted to let you know and make sure you don't think that we're not interested nor reacting fast enough! ;)

carlosame:
@vmassol thanks for the update. BTW XWiki's people are awesome in how fast you react.

This PR depends on css4j 3.0.0 which is not even released yet, so no problem about that. It is worth noting that when the XWiki's current PDF export uses my `DefaultEntityResolver`, that one is configured to block any outside connections (so it is even more restrictive than what this patch provides).

carlosame:
css4j 3.0.0 is now released. Once it is available in the repository, I can remove the draft status of this PR.

vmassol:
> Requires css4j 3.0 or higher

Note that XWiki is now using CSS4J 3.0.0 (https://jira.xwiki.org/browse/XWIKI-17898).

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/DefaultEntityResolver.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package org.xwiki.xml.internal;
+
+import javax.inject.Singleton;
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.xml.EntityResolver;
+
+/**
+ * Uses css4j's entity resolver.
```
vmassol, line 28: I think it would be good to explain why we use CSS4's resolver. I'd mention avoiding code duplication, and that it protects XWiki from most information leaks, DoS attacks based on entity expansion/recursion like the 'billion laughs attack' (copied from https://css4j.github.io/api/2.1/io/sf/carte/doc/xml/dtd/DefaultEntityResolver.html).

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/DefaultEntityResolver.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @version $Id$
+ */
+@Component
+@Singleton
+public class DefaultEntityResolver extends io.sf.carte.doc.xml.dtd.DefaultEntityResolver implements EntityResolver
+{
+    public DefaultEntityResolver() {
+        super(false);
+        registerSystemIdFilename("http://jackrabbit.apache.org/dtd/graffito-jcr-mapping.dtd",
```
vmassol, line 38: Could you explain why there's this here? It feels a bit strange that there's only 1 entry standalone here. I think it deserves some comments.

WDYT?

vmassol:
BTW I see you've left `LocalEntityResolver`. Why is that? Why would we still need it if we have `DefaultEntityResolver`?

vmassol:
Sorry for these questions but I'm not very familiar with this domain in XWiki and I'd like to understand before I can merge this PR. Thanks again for this PR! 👍

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/DefaultEntityResolver.java
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @version $Id$
+ */
+@Component
+@Singleton
+public class DefaultEntityResolver extends io.sf.carte.doc.xml.dtd.DefaultEntityResolver implements EntityResolver
+{
+    public DefaultEntityResolver() {
+        super(false);
+        // The following is used by org.apache.maven.wagon:wagon-webdav-jackrabbit
```
vmassol, line 50: Thanks for the comment but why is that needed? Is it because by default the resolver will only resolve DTDs locally and we don't provide locally the `graffito-jcr-mapping.dtd` needed by `wagon-webdav-jackrabbit`?  I'm curious to understand how it was working before your change. Also, shouldn't we provide it locally to speed it up and avoid the internet lookup? Actually I'm even wondering why we need wagon-webdav-jackrabbit as I doubt that we use it, so we might just want to exclude that dependency altogether (would need to check where it's used).

Thanks

xwiki-commons-core/xwiki-commons-xml/src/main/resources/META-INF/components.txt
```diff
@@ -7,5 +7,5 @@ org.xwiki.xml.internal.html.filter.ControlCharactersFilter
 org.xwiki.xml.internal.html.filter.AttributeFilter
 org.xwiki.xml.internal.html.filter.UniqueIdFilter
 org.xwiki.xml.internal.html.DefaultHTMLCleaner
-org.xwiki.xml.internal.LocalEntityResolver
```
vmassol, line 10: Since you removed it, it cannot be used anymore and thus I guess it means it's not used anymore in xwiki commons/rendering/platform. Thus we should be able to remove the `LocalEntityResolver` class altogether (or move it to a legacy module - https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HBackwardCompatibility).

@carlosame I could take over from there and do the work for the legacy move.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/DefaultEntityResolver.java
vmassol: > Actually I'm even wondering why we need wagon-webdav-jackrabbit as I doubt that we use it, so we might just want to exclude that dependency altogether (would need to check where it's used).

About this part, I can already confirm that we don't need it for XWiki at runtime since I cannot find it in `WEB-INF/lib`.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/DefaultEntityResolver.java
carlosame: Pushed a new version of the commit explaining why. I wasn't much verbose because you wanted some confidentiality (and Github PRs are quite visible).

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/DefaultEntityResolver.java
carlosame: A new version of the commit explains why. I looked for DTDs in all the XWiki repositories, finding the XHTML 1.0 `Transitional` and `Strict`, and the _graffito_ DTD. My resolver has all the XHTML 1.x DTDs built-in, so only the _graffito_ had to be dealt with.

It is worth mentioning that, as you did not include the XHTML 1.1 DTDs with XWiki, any attempt to use XHTML 1.1 would result in requests being sent to `w3.org` which is applying 15s delays to petitions:

```http
$ curl -I http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0  2611    0     0    0     0      0      0 --:--:--  0:00:15 --:--:--     0HTTP/1.1 200 OK
date: Thu, 01 Oct 2020 11:31:34 GMT
last-modified: Tue, 20 Mar 2018 05:54:52 GMT
cache-control: max-age=31536000
expires: Fri, 01 Oct 2021 11:31:34 GMT
vary: Accept-Encoding
strict-transport-security: max-age=15552000; includeSubdomains; preload
access-control-allow-origin: *
content-length: 2611
content-type: text/plain
etag: W/"**********************************"
accept-ranges: bytes
```
As a side-note, they aren't using a DTD `Content-Type`, instead they use `text/plain`. I have yet to see a DTD being served under a DTD type.

carlosame:
> I see you've left `LocalEntityResolver`. Why is that? Why would we still need it if we have `DefaultEntityResolver`?

In case the new resolver somehow breaks somebody's setup (imagine someone using `file:` system-ids to retrieve content from their internal filesystem), they can re-enable `LocalEntityResolver` in the `components.txt` configuration file while they devise a plan to upgrade their documents.

The `LocalEntityResolver` could be removed in version 13, for example.

xwiki-commons-core/xwiki-commons-xml/src/main/resources/META-INF/components.txt
carlosame: > it cannot be used anymore

Well it could be used again if the user is told to edit the `components.txt` file. See my other comment.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/DefaultEntityResolver.java
carlosame: From the `pom.xml`:

```xml
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-site-plugin</artifactId>
  <!-- Lock down plugin version for build reproducibility -->
  <version>3.3</version>
  <!-- Site Plugin doesn’t automatically pull the Apache Wagon Dav implementation which is needed for
	   site:deploy goal when using Dav protocol. -->
  <dependencies>
	<dependency>
	  <groupId>org.apache.maven.wagon</groupId>
	  <artifactId>wagon-webdav-jackrabbit</artifactId>
	  <version>2.6</version>
	</dependency>
  </dependencies>
</plugin>
```

carlosame: In any case, `maven-site-plugin` does not use XWiki's resolver, and a `fgrep --directories=recurse jackrabbit *` on XWiki's repositories does not yield anything, so the only reason to not remove the reference to that DTD would be if users are expected to use it in documents.

May I remove it?

vmassol: > In any case, maven-site-plugin does not use XWiki's resolver

I was about to say that :)

> May I remove it?

Yes, I don't see a need ATM.

Thanks

vmassol:
> In case the new resolver somehow breaks somebody's setup (imagine someone using file: system-ids to retrieve content from their internal filesystem), they can re-enable LocalEntityResolver in the components.txt configuration file while they devise a plan to upgrade their documents.

So, your change is actually breaking backward compatibility because you've modified the `default` hint for the `EntityResolver` role. No code is allowed to use the `LocalEntityResolver` class directly (it's in an `internal` package). If the two resolvers behave differently then we need to keep the previous one under the `default` hint and add a different hint for the new one (e.g. `local`). Maybe I'd change the name to be `CSS4JEntityResolver` for the implementation class name since `DefaultEntityResolver` is not very explicit.

FYI, users will use this to get it:

```
@Inject
@Name('local')
private EntityResolver resolver;
```

And right now they do the following to get `LocalEntityResolver`:

```
@Inject
private EntityResolver resolver;
```

WDYT?

vmassol:
> So, your change is actually breaking backward compatibility

Unless the 2 implementations have the same behaviors but from your comment, I think you're saying that they don't. Right?

carlosame:
> So, your change is actually breaking backward compatibility

Yes, this PR necessarily breaks backwards compatibility because the current behaviour is unsafe. Giving administrative rights on a Wiki should not imply ability to "read the server's filesystem with the credentials of the server", nor "access remote resources with the possible privileges of the server", nor "execute remote `jar:` bombs on the server". And that's what you are having right now.

You need to disable `LocalEntityResolver`, and only re-enable if there is a very compelling (and specific) need to do so.

carlosame:
The build failed with this:

```
[2020-10-05 12:27:55] [autobuild] [ERROR] Failed to execute goal fr.inria.gforge.spoon:spoon-maven-plugin:3.4.2:check (spoon-main) on project xwiki-commons-xml: Exception during the spoonify of the target project.: The following errors were found:
[2020-10-05 12:27:55] [autobuild] [ERROR] - Component [org.xwiki.xml.internal.LocalEntityResolver] is not declared in [/opt/src/xwiki-commons-core/xwiki-commons-xml/target/classes/META-INF/components.txt]! Consider adding it or if it is normal use the "staticRegistration" parameter as in "@Component(staticRegistration = false)"
```

I'm not familiar with _spoonification_, what should I do here, mark the `LocalEntityResolver` as `local`? (or `unsafe`?)

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/DefaultEntityResolver.java
vmassol: > It is worth mentioning that, as you did not include the XHTML 1.1 DTDs with XWiki,

That's a good point. We use HTML5 in most generated HTML so we probably don't need it. But CKEditor still uses XHTML 1.1 I think so it's possible that we're having extra delays there because of it.

vmassol:
> I'm not familiar with _spoonification_, what should I do here, mark the `LocalEntityResolver` as `local`? (or `unsafe`?)

This is just what I was telling you. You cannot remove the `LocalEntityResolver` component since you maybe break any code using it (extensions for example which can be installed at runtime in XWiki). This is in essence what the build is telling you.

So we need to decide if we want to voluntarily break extensions or not.

Case 1 (we don't want to break backward compat)
* Keep the `LocalEntityResolver` component (but move it to legacy) and add a new one with a different hint. Maybe add a log in `LocalEntityResolver` to mention that it should no longer be used.

Case 2 (we're ok to break backward compat because we think it's too important from a security POV):
* Remove the `LocalEntityResolver` class and replace it with the `CSS4JEntityResolver` one, using the same hint. This can break extensions using the resolver in ways that result in different behaviors.

@tmortagne @surli any opinions?

carlosame:
As a side note, `HardenedEntityResolver` sounds better than `CSS4JEntityResolver`. After all, this resolver was a late addition to CSS4J (used to be part of JCLF).

vmassol:
> As a side note, `HardenedEntityResolver` sounds better than `CSS4JEntityResolver`

The problem with that type of generic name is that anyone can claim it. Imagine we have a next one that is even more hardened. Would it be called `EvenMoreHardenedEntityResolver`? :)  Since this is the impl class, I usually prefer to name it after what's specific in the implementation.

vmassol:
> > I'm not familiar with _spoonification_, what should I do here, mark the `LocalEntityResolver` as `local`? (or `unsafe`?)
> 
> This is just what I was telling you. You cannot remove the `LocalEntityResolver` component since you maybe break any code using it (extensions for example which can be installed at runtime in XWiki). This is in essence what the build is telling you.
> 
> So we need to decide if we want to voluntarily break extensions or not.
> 
> Case 1 (we don't want to break backward compat)
> 
> * Keep the `LocalEntityResolver` component (but move it to legacy) and add a new one with a different hint. Maybe add a log in `LocalEntityResolver` to mention that it should no longer be used.
> 
> Case 2 (we're ok to break backward compat because we think it's too important from a security POV):
> 
> * Remove the `LocalEntityResolver` class and replace it with the `CSS4JEntityResolver` one, using the same hint. This can break extensions using the resolver in ways that result in different behaviors.
> 
> @tmortagne @surli any opinions?

After discussing with Simon we both agree that Case 2 is ok.

tmortagne:
> @tmortagne @surli any opinions?

We can get rid of `LocalEntityResolver` completely (Case 2) IMO.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/DefaultEntityResolver.java
carlosame: > We use HTML5 in most generated HTML

If you process that HTML5 with a normal XML toolchain (I mean with an XML parser), you need a resolver like mine which implements `EntityResolver2`. The PDF export process (which already uses the resolver) can resolve the entities thanks to that.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/DefaultEntityResolver.java
carlosame: Removed.

vmassol:
@carlosame sounds good to me! thanks! 

Do you know if I can apply it as is or do I also need to apply some other PRs you did for platform at the same time since the behavior changed a bit? I know you did some other PRs but I haven't reviewed them yet and don't know if they're needed at the same time or not. Thx

carlosame:
> Do you know if I can apply it as is or do I also need to apply some other PRs

You can apply this PR independently of the others.

xwiki-commons-core/xwiki-commons-xml/pom.xml
```diff
@@ -78,6 +78,11 @@
       <groupId>com.fasterxml.woodstox</groupId>
       <artifactId>woodstox-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.sf.carte</groupId>
+      <artifactId>xml-dtd</artifactId>
+      <version>3.0.0</version>
+    </dependency>
```
vmassol, line 85: @carlosame Hi. I didn't realized it back then when I merged your PR but we cannot do this since this artifact is currently NOT on Maven Central and all our deps inside xwiki-commons and xwiki-rendering must be on Maven Central since we deploy all these modules to Maven Central and they need to be self-standing (that's one rule of Maven Central).

Is there any possibility that you could deploy this artifact to Maven Central (through Sonatype's nexus repo as we do for ex)?

If not, I fear we'll have to rollback this PR and that would be a shame IMO.

Thanks a lot

carlosame: Hi,

Unfortunately my concerns explained in:

https://groups.google.com/g/css4j/c/op5jIoINb3M/m/IiiN-LfkDAAJ

do apply. There is no way that my artifacts are going into Central.

PR #109, opened 2020-10-16
XCOMMONS-2042: Add a method in URLTool to convert a string in an URL object

ignatescustefan:
The method should return an URL object from a given string if the string is right formatted, otherwise null.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/URLTool.java
+++ b/xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/URLTool.java
@@ -66,7 +66,7 @@ public class URLTool
      *
      * @param urlString string to be convert
      * @return an URL object from a given string
-     * @since 12.9RC
+     * @since 12.9RC1
      */
     public URL toURL(String urlString)
     {
--- a/xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/URLTool.java
+++ b/xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/URLTool.java
@@ -19,6 +19,8 @@
  */
 package org.xwiki.velocity.tools;
 
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
@@ -57,5 +59,20 @@ public class URLTool
             }
         }
         return queryParams;
+    }
+
+    /**
+     * Convert a string to an URL object, if it's possible.
+     *
+     * @param urlString string to be convert
+     * @return an URL object from a given string
+     */
+    public URL getURL(String urlString)
+    {
+        try {
+            return new URL(urlString);
+        } catch (MalformedURLException e) {
+            return null;
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


xwiki-commons-core/xwiki-commons-velocity/src/test/java/org/xwiki/velocity/tools/URLToolTest.java
```diff
@@ -81,4 +84,18 @@ void preserveParameterOrder()
         String queryString = "x=5&x=4&r=3&a=2";
         assertEquals(queryString, escapeTool.url(tool.parseQuery(queryString)));
     }
+
+    @Test
+    void parseValidURL() throws MalformedURLException
+    {
+        URL url = tool.getURL("http://www.xwiki.com");
+        assertEquals(url, new URL("http://www.xwiki.com"));
```
vmassol, line 92: wrong order.

vmassol: I'd suggest `getURLValid` for the test name to follow our best practices.

xwiki-commons-core/xwiki-commons-velocity/src/test/java/org/xwiki/velocity/tools/URLToolTest.java
```diff
@@ -81,4 +84,18 @@ void preserveParameterOrder()
...
+    }
+
+    @Test
+    void parseInvalidURL()
```
vmassol, line 96: Same here, I'd suggest: `getURLInvalid`

vmassol:
Another idea is to have a String to URL converter (in xwiki-properties). I'm surprised that this doesn't exist already. But maybe you don't need to pass a URL to a java method?

tmortagne:
> Another idea is to have a String to URL converter (in xwiki-properties). I'm surprised that this doesn't exist already. But maybe you don't need to pass a URL to a java method?

Such converter does exist but that's not @ignatescustefan's use case (question already asked on the jira issue :)).

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/URLTool.java
```diff
@@ -58,4 +60,19 @@
...
         return queryParams;
     }
+
+    /**
+     * Convert a string to an URL object, if it's possible.
+     *
+     * @param urlString string to be convert
+     * @return an URL object from a given string
+     */
+    public URL getURL(String urlString)
```
tmortagne, line 70: I feel like the method should be called `parseURL` since that's what it's really about.

xwiki-commons-core/xwiki-commons-velocity/src/test/java/org/xwiki/velocity/tools/URLToolTest.java
```diff
@@ -81,4 +84,18 @@ void preserveParameterOrder()
         String queryString = "x=5&x=4&r=3&a=2";
         assertEquals(queryString, escapeTool.url(tool.parseQuery(queryString)));
     }
+
+    @Test
+    void getURLValid() throws MalformedURLException
```
tmortagne, line 89: As suggested by @vmassol in a previous comment, the name of the test method should use the name of the tested method.

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/URLTool.java
```diff
@@ -58,4 +60,20 @@
         }
         return queryParams;
     }
+
+    /**
+     * Convert a string to an URL object, if it's possible.
+     *
+     * @param urlString string to be convert
+     * @return an URL object from a given string
+     * @since 12.9RC
```
tmortagne, line 69: "12.9RC1"

tmortagne:

PR #124, opened 2020-12-10
XCOMMONS-2099 : Allow to define HTTP headers to Maven extension repositories

aubincleme:
Add support for HTTP headers in the extension repository configuration for Maven repositories


xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-repositories/xwiki-commons-extension-repository-maven/src/main/java/org/xwiki/extension/repository/aether/internal/AetherExtensionRepository.java
```diff
@@ -242,6 +245,18 @@ protected XWikiRepositorySystemSession createRepositorySystemSession() throws Re
             throw new ResolveException("Failed to create the repository system session", e);
         }
 
+        // Add HTTP headers separately from the rest of the configuration of the session as they have to be
+        // stored as a map in the session configuration properties.
+        Map<String, String> sessionHttpHeaders = new HashMap<>();
+        for (Map.Entry<String, String> property : getDescriptor().getProperties().entrySet()) {
+            if (StringUtils.startsWith(property.getKey(), "httpHeaders.")) {
```
tmortagne, line 252: I would use "http.headers." as prefix since we might have other http related properties in the future.

PR #133, opened 2021-02-15
XCOMMONS-2068: Possible OutOfBoundsException when performing diff

surli:
* Bulletproof a bit more the DefaultUnifiedDiffDisplayer to prevent an
    OutOfBoundException

### DIFF
--- a/xwiki-commons-core/xwiki-commons-diff/xwiki-commons-diff-display/src/main/java/org/xwiki/diff/display/internal/DefaultUnifiedDiffDisplayer.java
+++ b/xwiki-commons-core/xwiki-commons-diff/xwiki-commons-diff-display/src/main/java/org/xwiki/diff/display/internal/DefaultUnifiedDiffDisplayer.java
@@ -426,7 +426,7 @@ public class DefaultUnifiedDiffDisplayer implements UnifiedDiffDisplayer
         // Add the unmodified elements before the given delta.
         int count = state.getBlocks().peek().isEmpty() ? contextSize : contextSize * 2;
         int lastChangeIndex = state.getLastDelta() == null ? -1 : state.getLastDelta().getPrevious().getLastIndex();
-        int end = delta.getPrevious().getIndex();
+        int end = Math.min(delta.getPrevious().getIndex(), state.getPrevious().size());
         int start = Math.max(end - count, lastChangeIndex + 1);
         state.getBlocks().peek().addAll(this.<E, F>getUnmodifiedElements(state.getPrevious(), start, end));
 
@@ -500,7 +500,9 @@ public class DefaultUnifiedDiffDisplayer implements UnifiedDiffDisplayer
     private <E, F> List<UnifiedDiffElement<E, F>> getUnmodifiedElements(List<E> previous, int start, int end)
     {
         List<UnifiedDiffElement<E, F>> unmodifiedElements = new ArrayList<>();
-        for (int i = start; i < end; i++) {
+        // Prevent exceeding the size of the previous list in the loop.
+        int realEnd = Math.min(end, previous.size());
+        for (int i = start; i < realEnd; i++) {
             unmodifiedElements.add(new UnifiedDiffElement<>(i, Type.CONTEXT, previous.get(i)));
         }
         return unmodifiedElements;


xwiki-commons-core/xwiki-commons-diff/xwiki-commons-diff-display/src/main/java/org/xwiki/diff/display/internal/DefaultUnifiedDiffDisplayer.java
```diff
@@ -426,7 +426,7 @@ public void setLastConflict(Conflict<E> lastConflict)
         // Add the unmodified elements before the given delta.
         int count = state.getBlocks().peek().isEmpty() ? contextSize : contextSize * 2;
         int lastChangeIndex = state.getLastDelta() == null ? -1 : state.getLastDelta().getPrevious().getLastIndex();
-        int end = delta.getPrevious().getIndex();
+        int end = Math.min(delta.getPrevious().getIndex(), state.getPrevious().size());
```
mflorea, line 429: I don't like this. If ``delta.previous.index`` is greater than ``previous.size`` then it means the delta is wrong and the problem is else where. You're hiding a bug in the way the delta was computed. I don't think this code should handle bad deltas. You can throw an exception is the delta is not valid, but I'm not sure this is the right place.

xwiki-commons-core/xwiki-commons-diff/xwiki-commons-diff-display/src/main/java/org/xwiki/diff/display/internal/DefaultUnifiedDiffDisplayer.java
```diff
@@ -500,7 +500,9 @@ public void setLastConflict(Conflict<E> lastConflict)
     private <E, F> List<UnifiedDiffElement<E, F>> getUnmodifiedElements(List<E> previous, int start, int end)
     {
         List<UnifiedDiffElement<E, F>> unmodifiedElements = new ArrayList<>();
-        for (int i = start; i < end; i++) {
+        // Prevent exceeding the size of the previous list in the loop.
+        int realEnd = Math.min(end, previous.size());
+        for (int i = start; i < realEnd; i++) {
```
mflorea, line 505: What if ``start`` is negative? I don't think this method should validate the start and end points. Again, I'm worried that you're going to hide a real problem that is elsewhere.

xwiki-commons-core/xwiki-commons-diff/xwiki-commons-diff-display/src/main/java/org/xwiki/diff/display/internal/DefaultUnifiedDiffDisplayer.java
surli: > You're hiding a bug in the way the delta was computed.

I'm mixed here. I agree that I'm not fixing a bug about the way deltas are built if the `delta.previous.index` is greater than `previous.size`. Now the issue here is that I've not been able to reproduce such bug, so I thought it was better to protect the code against a potential failure, even if it would display wrong result, than to let it crash. 
Re exceptions, I checked again the code and apparently we never throw any exception in those methods. That could be probably improved but would be a possible breaking change. 
What I might do right now is to add an error logger if such bad delta happens again.

surli: Hmm I just checked the jacoco report and apparently there's a part of my code which is not covered by the tests and which might lead to such bug. Here: https://github.com/xwiki/xwiki-commons/blob/master/xwiki-commons-core/xwiki-commons-diff/xwiki-commons-diff-display/src/main/java/org/xwiki/diff/display/internal/DefaultUnifiedDiffDisplayer.java#L235-L239

I need to find a scenario to cover this part in my test.

surli: For info I created this small script to generate files with potential conflict: https://gist.github.com/surli/5506e2427daa4cad6279fbfea8f7f172 so far I've been able to find another bug in DefaultDiffManager#merge but I haven't managed to expose this one even after generating ~20 variations.

surli:
Closing this one in favor of #156

PR #139, opened 2021-04-01
XCOMMONS-2116: Upgrade to Infinispan 12.0.1

Jobkanis:
Issue: https://jira.xwiki.org/projects/XCOMMONS/issues/XCOMMONS-2116

Upgrade Infinispan to version 12.0.1.Final as found in the issue.
This goes together with an upgrade of org.infinispan.protostream as mentioned in https://infinispan.org/docs/stable/titles/upgrading/upgrading.html.

Also xwiki-commons-core/xwiki-commons-cache/xwiki-commons-cache-infinispan/src/test/java/org/xwiki/cache/infinispan/InfinispanConfigTest.java kept failing due to an incorrect string compare. This is now fixed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1384,7 +1384,7 @@
       <dependency>
         <groupId>org.infinispan</groupId>
         <artifactId>infinispan-core</artifactId>
-        <version>11.0.10.Final</version>
+        <version>12.0.1.Final</version>
         <exclusions>
           <!-- We use jakarta.transaction:jakarta.transaction-api (the reference API id) instead in XWiki -->
           <exclusion>
--- a/pom.xml
+++ b/pom.xml
@@ -1379,7 +1379,7 @@
       <dependency>
         <groupId>org.infinispan.protostream</groupId>
         <artifactId>protostream</artifactId>
-        <version>4.3.5.Final</version>
+        <version>4.4.0.Beta2</version>
       </dependency>
       <dependency>
         <groupId>org.infinispan</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -1376,6 +1376,7 @@
         <artifactId>protobuf-java</artifactId>
         <version>3.15.6</version>
       </dependency>
+      <!-- Protostream 4.4.0 is required by Infinispan 12.0.2. Since no final version of Protostream is available, the Beta2 version will be used. -->
       <dependency>
         <groupId>org.infinispan.protostream</groupId>
         <artifactId>protostream</artifactId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -1379,12 +1379,12 @@
       <dependency>
         <groupId>org.infinispan.protostream</groupId>
         <artifactId>protostream</artifactId>
-        <version>4.3.5.Final</version>
+        <version>4.4.0.Beta2</version>
```
tmortagne, line 1382: It's not clear to me why we need a beta version. The minimum would be to put a comment to explain it.

pom.xml
```diff
@@ -1379,12 +1379,12 @@
...
       </dependency>
       <dependency>
         <groupId>org.infinispan</groupId>
         <artifactId>infinispan-core</artifactId>
-        <version>11.0.10.Final</version>
+        <version>12.0.1.Final</version>
```
tmortagne, line 1387: The last version is 12.0.2 so it does not make much sense to upgrade to 12.0.1.

xwiki-commons-core/xwiki-commons-cache/xwiki-commons-cache-infinispan/src/test/java/org/xwiki/cache/infinispan/InfinispanConfigTest.java
```diff
@@ -115,7 +115,7 @@ void createCacheWhenFileCacheNoPath() throws Exception
         assertEquals(1, stores.size());
 
         String location = ((SingleFileStoreConfiguration) stores.get(0)).location();
-        assertEquals(this.environment.getTemporaryDirectory().toString() + "/cache", location);
+        assertEquals(this.environment.getTemporaryDirectory().toString() + "\\cache", location);
```
tmortagne, line 118: You seem to have converted a test that was based on UNIX separator ("/") to Windows separator ("\"), this means it's going to fail everywhere else than Windows.

pom.xml
Jobkanis: Infinispan 12 requires version 4.4.0 of protostream, which is, as far as I know, only available as Alpha or Beta version.

Jobkanis: I will add a comment to explain it.

pom.xml
Jobkanis: I have changed the issue and will be looking into upgrading to version 12.0.2.

xwiki-commons-core/xwiki-commons-cache/xwiki-commons-cache-infinispan/src/test/java/org/xwiki/cache/infinispan/InfinispanConfigTest.java
Jobkanis: I will revert this change. Thanks for noticing.

pom.xml
tmortagne: I think I would prefer to wait until protostream 4.4 final is out. It's not like we were is a hurry to upgrade Infinispan.

tmortagne:
Implemented as part of https://github.com/xwiki/xwiki-commons/commit/8106210299a0df4ab5595d1a82c2cfb353e97e23.

PR #141, opened 2021-04-06
Remove redundant dependencies

Celebrate-future:
@tmortagne Hi, I am a user of project **_org.xwiki.commons:xwiki-commons-tool-remote-resource-plugin:13.3-SNAPSHOT_**. I found that its pom file introduced **_55_** dependencies. However, among them, **_11_** libraries (**_20%_**) have not been used by your project (the redundant dependencies are listed below). Reduce these useless dependencies can help prevent conflicts between library versions. MeanWhile, it can minimize the total added size to projects. It can also help enable advanced scenarios for users of your package. 
This PR helps **_org.xwiki.commons:xwiki-commons-tool-remote-resource-plugin:13.3-SNAPSHOT_** lose weight :) I have tested the revised configuration in my local environment. It is safe to remove the unused libraries.

Best regards

## Redundant dependencies----
<pre><code>
org.sonatype.plexus:plexus-sec-dispatcher:jar:1.4:compile
org.sonatype.plexus:plexus-cipher:jar:1.4:compile
javax.enterprise:cdi-api:jar:1.0:compile
javax.inject:javax.inject:jar:1:compile
org.apache.maven:maven-archiver:jar:3.5.0:compile
org.codehaus.plexus:plexus-archiver:jar:4.2.4:compile
org.codehaus.plexus:plexus-io:jar:3.2.0:compile
org.apache.commons:commons-compress:jar:1.20:compile
org.iq80.snappy:snappy:jar:0.4:compile
org.tukaani:xz:jar:1.8:runtime
org.codehaus.plexus:plexus-resources:jar:1.0.1:compile
</code></pre>

### DIFF
--- a/xwiki-commons-tools/xwiki-commons-tool-remote-resource-plugin/pom.xml
+++ b/xwiki-commons-tools/xwiki-commons-tool-remote-resource-plugin/pom.xml
@@ -45,6 +45,28 @@
     <dependency>
       <groupId>org.apache.maven.plugins</groupId>
       <artifactId>maven-remote-resources-plugin</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>org.sonatype.plexus</groupId>
+          <artifactId>plexus-sec-dispatcher</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.codehaus.plexus</groupId>
+          <artifactId>plexus-resources</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>javax.inject</groupId>
+          <artifactId>javax.inject</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>javax.enterprise</groupId>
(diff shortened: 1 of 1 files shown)


tmortagne:
I'm generally not a fan of this kind of optimization. Even if the excluded dependencies are not used they don't really cause any trouble right now while we are not really sure of what will happen in some edge cases if we do exclude it and it could cause unnoticed regressions when upgrading maven-remote-resources-plugin and it start actually using one of those dependencies.

In any case I cannot really validate if this pull request is accurate since I don't know maven-remote-resources-plugin well enough. We just extend this plugin to workaround a memory problem in ProcessRemoteResourcesMojo (see https://issues.apache.org/jira/browse/MRRESOURCES-106).

vmassol:
Yes I also don't like it too much. Using an exclusion is usually a hack, and kind of suggests that there's a problem in the upstream code that it's missing some "optional" deps.

PR #142, opened 2021-04-13
XCOMMONS-2188: NPE when interacting with the Infinispan Statistics component via JMX

slauriere:


xwiki-commons-core/xwiki-commons-cache/xwiki-commons-cache-infinispan/pom.xml
```diff
@@ -50,6 +50,11 @@
       <groupId>org.infinispan</groupId>
       <artifactId>infinispan-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.microprofile.metrics</groupId>
+      <artifactId>microprofile-metrics-api</artifactId>
+      <version>3.0</version>
+    </dependency>
```
vmassol, line 57: Should this be optional?

vmassol: I'd also add some comment explaining this dep

slauriere: Thanks Vincent for the review. I realize that the dependency should actually be added only to the modules needing such a JMX access, so this is the wrong place I guess, apologies. Let me investigate a bit further and probably I will close the PR and the related issue.

slauriere:
Closing as invalid: the optional dependency should be pulled only by the projects needing JMX access to the Infinispan Statistics component.

xwiki-commons-core/xwiki-commons-cache/xwiki-commons-cache-infinispan/pom.xml
tmortagne: No I think this is the right place. Just missing a comment explaining why we need this dependency now.

vmassol: I agree that it would be nice to be able to use JMX to access infinispan stats by default in XS. I'm not sure the location is right since IMO it could also be in the minimal dependencies module in platform (for the base flavor).

Is this a regression due to an upgrade of infinispan? I'm positive it was working before. If it's a regression I would mark XCOMMONS-2188 as such since I believe we can consider JMX access as base feature of XS.

slauriere: Yes it's a regression due to an upgrade of Infinispan which introduced this dependency in the 10.x lifecycle, I mentioned the related Infinispan commits / issues in XCOMMONS-2188, which Thomas marked as a regression indeed in the meantime.

tmortagne:
Reopenning since the pull request is valid, just missing some improvements.

xwiki-commons-core/xwiki-commons-cache/xwiki-commons-cache-infinispan/pom.xml
```diff
@@ -50,6 +50,11 @@
       <groupId>org.infinispan</groupId>
       <artifactId>infinispan-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.microprofile.metrics</groupId>
+      <artifactId>microprofile-metrics-api</artifactId>
+      <version>3.0</version>
```
tmortagne, line 56: I think I would put this version in a dependencyManagement along with the Infinispan one in the xwiki-commons root pom.

vmassol: Regarding commons, I don't remember why we put platform-related depMgmt there. At first sight it doesn't look legit and should be put in platform's pom.xml.

tmortagne: >  I don't remember why we put platform-related depMgmt there

We don't.

vmassol: Then we shouldn't put the dep there then, since it's related to distribution wanting to bundle support for JMX by default.

tmortagne: I don't understand what you mean. This is a dependency of xwiki-commons-cache-infinispan here, so I'm suggesting to move the version to xwiki-common pom.

If your goal was to suggest moving this dependency entirely to xwiki-platform then that's a different subject.

vmassol: I don't think it's a mandatory dep of xwiki-commons-cache-infinispan since we already a dep on infinispan core and it's not bringing it, which most likely means it's an optional dep. Right?

tmortagne: It all depends if you want xwiki-commons-cache-infinispan to properly expose Infinispan JMX (which always been the case before) or not. Right now the regression is at xwiki-commons-cache-infinispan level.

vmassol: Yeah depends if we want to make the minimal flavor fatter or not I guess...

tmortagne: > Yeah depends if we want to make the minimal flavor fatter or not I guess...

That dependency have to be in the WAR for Infinispan to find it anyway so it's not like putting it on platform side was going to change anything to the XWiki WAR size. Also we are talking about 34kB here.

vmassol: I was not referring to the XS WAR but to the minimal one which could be smaller with some optional features that not everyone might need.

vmassol: I'm +0 to add it to commons, not blocking it, just trying to think if it's the right thing to do.

slauriere: I moved the dependency declaration to xwiki-commons with the correct version and a comment. As for unit testing: I checked a bit how to proceed, not sure I'm able to provide a working one on my own for now though, any guidance or example welcome or I'll check how you did it.

slauriere: (moved the dependency in case it can be useful indeed despite not knowing either what's best)

vmassol:
I would write a functional test to test that JMX is working (instead of a unit test), since we don't have a single JMX func test AFAIK.

slauriere:
Right, I see, I restored the dependency in xwiki-commons-cache-infinispan without the version, and kept version 2.3 in the root component for now since I'm not sure how to evaluate quickly the scope of changes between 2.3 and 3.0.

tmortagne:
> I'm not sure how to evaluate quickly the scope of changes between 2.3 and 3.0.

At least you should use 2.3.3.

slauriere:
> At least you should use 2.3.3.

Just did that, I was not entirely sure since Infinispan master branch still uses version 2.3 as far as I can see but I understand it's probably safe and better indeed.

tmortagne:
One last thing @slauriere, it should be a runtime scope dependency (in `xwiki-commons-cache-infinispan`) if we want to be fully accurate.

slauriere:
Scope added, thank you @tmortagne and @vmassol.

tmortagne:
Thanks for the work and the debugging @slauriere !

PR #147, opened 2021-09-06
[Misc] Update TestEnvironment#getResource to only resolve resources local to the tested module

manuelleduc:


vmassol:
Hi Manue.

Typo in the title: "locale" -> "local" :)

Could you explain the rationale for it? I don't see it in the comments in the code. Aren't there use cases where we want to resolve resources located in dependent jars? I can think of several tests that do that. For example for PageTest and testing templates we add a dep on platform-web-templates for this reason.

Thx

tmortagne:
> For example for PageTest and testing templates we add a dep on platform-web-templates for this reason.

The template manager do get templates from the classloader, right after trying in the Environment. It would not work in XWiki Standard otherwise.

vmassol:
> The template manager do get templates from the classloader, right after trying in the Environment. It would not work in XWiki Standard otherwise.

Then the PR title is wrong... That's what I was commenting about.

manuelleduc:
The proposed change is too generic and can cause regressions in other tests, I'll open another PR with a more suited change.

PR #148, opened 2021-09-07
XCOMMONS-2272: ERROR: 'Length limit reached'

ClemensRobbenhaar:
add custom error listener to overwrite the default error listener which fills the standard output with expected error messages

I did not write unit tests for this, thus the slightly lower coverage.
I feel it to be relatively hard to write a unit test to check this and also e.g. the `warning` method on the error listener might never be called from our code but must be proviced to implement the interface. I hope this is acceptable for a minor issue.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
@@ -27,6 +27,7 @@ import java.util.regex.Pattern;
 
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.ErrorListener;
 import javax.xml.transform.OutputKeys;
 import javax.xml.transform.Result;
 import javax.xml.transform.Source;
@@ -63,6 +64,59 @@ import org.xwiki.stability.Unstable;
  */
 public final class XMLUtils
 {
+    /**
+     * A simple error logging helper to suppress expected error messages.
+     * While extracting text excerpts via the {@link #extractXML} method
+     * the extraction throws expected exceptions eg. in case a text node
+     * is too long. The default ErrorListener writes these errors to the console.
+     * However as these errors are expected under normal conditions
+     * and are either handled in the {@link #extractXML} method
+     * or passed up in the chain of commands it is ok to log these
+     * at a much lower level, where users normally never see them.
+     */
+    private static class RelaxedErrorListener implements ErrorListener
+    {
+        private static final String STACK_TRACE_NOTE = "stack trace for information only";
+
+        /**
+         * Fatal errors are unexpected and are logged as warnings here.
+         * They are not really fatal to XWiki but should be handled
+         * up the command chain later, or maybe propagated to the UI level.
+         */
+        @Override
+        public void fatalError(TransformerException exception) throws TransformerException
+        {
+            LOGGER.warn("fatal error from xml transformer", exception);
+        }
+
+        /**
+         * Errors are expected and are only logged as debug.
+         * These errors happen e.g. if a text node exceeds the expected
+         * maximal length.
+         */
+        @Override
+        public void error(TransformerException exception) throws TransformerException
+        {
+            LOGGER.debug("error [{}] from xml transformer", exception.getMessage());
+            if (LOGGER.isTraceEnabled()) {
+                LOGGER.trace(STACK_TRACE_NOTE, exception);
+            }
+        }
+
+        /**
+         * Warnings are logged at debug level.
+         * They might be of concern for the developers but not the end users.
+         */
+        @Override
+        public void warning(TransformerException exception) throws TransformerException
+        {
+            LOGGER.debug("warning [{}] from xml transformer", exception.getMessage());
+            if (LOGGER.isTraceEnabled()) {
+                LOGGER.trace(STACK_TRACE_NOTE, exception);
+            }
+        }
+    }
+
     /** Logging helper object. */
     private static final Logger LOGGER = LoggerFactory.getLogger(XMLUtils.class);
 
@@ -110,6 +164,9 @@ public final class XMLUtils
     /** Xerces configuration parameter for disabling fetching and checking XMLs against their DTD. */
     private static final String DISABLE_DTD_PARAM = "http://apache.org/xml/features/nonvalidating/load-external-dtd";
 
+    /** Helper to log expected errors at an appropriate level. */
+    private static final ErrorListener RELAXED_ERROR_LISTENER = new RelaxedErrorListener();
+
     static {
         DOMImplementationLS implementation = null;
         try {
@@ -145,6 +202,7 @@ public final class XMLUtils
         try {
             handler = new ExtractHandler(start, length);
             Transformer xformer = TransformerFactory.newInstance().newTransformer();
+            xformer.setErrorListener(RELAXED_ERROR_LISTENER);
             xformer.transform(new DOMSource(node), new SAXResult(handler));
             return handler.getResult();
         } catch (Throwable t) {
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
@@ -90,15 +90,13 @@ public final class XMLUtils
         public void fatalError(TransformerException exception) throws TransformerException
         {
             LOGGER.warn("Fatal error from xml transformer: [{}]", ExceptionUtils.getRootCauseMessage(exception));
-            if (LOGGER.isTraceEnabled()) {
-                LOGGER.trace(STACK_TRACE_NOTE, exception);
-            }
+            LOGGER.trace(STACK_TRACE_NOTE, exception);
         }
 
         /**
-         * Errors are expected and are only logged as debug
+         * Errors are expected and are only logged as debug.
          * These errors happen e.g. if a text node exceeds the expected
-         * maximal length.
+         * maximal length, which can happen in regular usage.
          *
          * @param exception the exception to be logged
          */
@@ -106,9 +104,7 @@ public final class XMLUtils
         public void error(TransformerException exception) throws TransformerException
         {
             LOGGER.debug("Error [{}] from xml transformer", ExceptionUtils.getRootCauseMessage(exception));
-            if (LOGGER.isTraceEnabled()) {
-                LOGGER.trace(STACK_TRACE_NOTE, exception);
-            }
+            LOGGER.trace(STACK_TRACE_NOTE, exception);
         }
 
         /**
@@ -121,9 +117,7 @@ public final class XMLUtils
         public void warning(TransformerException exception) throws TransformerException
         {
             LOGGER.debug("Warning [{}] from xml transformer", ExceptionUtils.getRootCauseMessage(exception));
-            if (LOGGER.isTraceEnabled()) {
-                LOGGER.trace(STACK_TRACE_NOTE, exception);
-            }
+            LOGGER.trace(STACK_TRACE_NOTE, exception);
         }
     }
 
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
@@ -41,6 +41,7 @@ import javax.xml.transform.sax.SAXSource;
 import javax.xml.transform.stream.StreamResult;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.w3c.dom.Document;
@@ -88,7 +89,7 @@ public final class XMLUtils
         @Override
         public void fatalError(TransformerException exception) throws TransformerException
         {
-            LOGGER.warn("Fatal error from xml transformer: ", getRootMessage(exception));
+            LOGGER.warn("Fatal error from xml transformer: [{}]", ExceptionUtils.getRootCauseMessage(exception));
             if (LOGGER.isTraceEnabled()) {
                 LOGGER.trace(STACK_TRACE_NOTE, exception);
             }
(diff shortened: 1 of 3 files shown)


xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -63,6 +64,59 @@
...
+
+        /**
+         * Fatal errors are unexpected and are logged as warnings here.
+         * They are not really fatal to XWiki but should be handled
+         * up the command chain later, or maybe propagated to the UI level.
+         */
+        @Override
+        public void fatalError(TransformerException exception) throws TransformerException
+        {
+            LOGGER.warn("fatal error from xml transformer", exception);
```
vmassol, line 89: Note that we log the full stack trace only for errors, see https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HLoggingBestPractices

So either it's a warning or an error, we/you need to choose.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -63,6 +64,59 @@
  */
 public final class XMLUtils
 {
+    /**
+     * A simple error logging helper to suppress expected error messages.
```
vmassol, line 68: What is an "expected error message"? It's not clear from this javadoc.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
ClemensRobbenhaar: it is a warning. I will change the stack trace to "on trace only" as in the other methods.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -63,6 +64,59 @@
...
+     * While extracting text excerpts via the {@link #extractXML} method
+     * the extraction throws expected exceptions eg. in case a text node
+     * is too long. The default ErrorListener writes these errors to the console.
+     * However as these errors are expected under normal conditions
+     * and are either handled in the {@link #extractXML} method
+     * or passed up in the chain of commands it is ok to log these
+     * at a much lower level, where users normally never see them.
```
vmassol, line 75: I don't understand fully this comment. Since you log the error it'll be seen in the logs by users. It seems this is just replacing the XML parser logging with an xwiki logging. Right? Basically, I don't see how this suppresses error messages. Thanks!

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -63,6 +64,59 @@
...
+         * up the command chain later, or maybe propagated to the UI level.
+         */
+        @Override
+        public void fatalError(TransformerException exception) throws TransformerException
+        {
+            LOGGER.warn("fatal error from xml transformer", exception);
+        }
+
+        /**
+         * Errors are expected and are only logged as debug.
```
vmassol, line 93: Why are errors expected?

vmassol: Ok I think I understand now. You want to swallow "ERROR: 'Length limit reached'". I'm wondering if this is true for all errors, this makes them basically invisible and thus unknown.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -63,6 +64,59 @@
...
+         * These errors happen e.g. if a text node exceeds the expected
+         * maximal length.
+         */
+        @Override
+        public void error(TransformerException exception) throws TransformerException
+        {
+            LOGGER.debug("error [{}] from xml transformer", exception.getMessage());
```
vmassol, line 100: Should be un uppercase for `error` since it starts a sentence. Also I would log the full stack trace or at least the root message (but not the last message which is not always useful).

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -63,6 +64,59 @@
...
+        }
+
+        /**
+         * Warnings are logged at debug level.
+         * They might be of concern for the developers but not the end users.
+         */
+        @Override
+        public void warning(TransformerException exception) throws TransformerException
+        {
+            LOGGER.debug("warning [{}] from xml transformer", exception.getMessage());
```
vmassol, line 113: Same comment re the message.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -63,6 +64,59 @@
...
+        {
+            LOGGER.debug("error [{}] from xml transformer", exception.getMessage());
+            if (LOGGER.isTraceEnabled()) {
+                LOGGER.trace(STACK_TRACE_NOTE, exception);
+            }
+        }
+
+        /**
+         * Warnings are logged at debug level.
+         * They might be of concern for the developers but not the end users.
```
vmassol, line 108: Missing the javadoc for the parameter, that should trigger checkstyle normally

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -63,6 +64,59 @@
...
+        @Override
+        public void fatalError(TransformerException exception) throws TransformerException
+        {
+            LOGGER.warn("fatal error from xml transformer", exception);
+        }
+
+        /**
+         * Errors are expected and are only logged as debug.
+         * These errors happen e.g. if a text node exceeds the expected
+         * maximal length.
```
vmassol, line 95: Missing the javadoc for the parameter, that should trigger checkstyle normally

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -63,6 +64,59 @@
...
+     */
+    private static class RelaxedErrorListener implements ErrorListener
+    {
+        private static final String STACK_TRACE_NOTE = "stack trace for information only";
+
+        /**
+         * Fatal errors are unexpected and are logged as warnings here.
+         * They are not really fatal to XWiki but should be handled
+         * up the command chain later, or maybe propagated to the UI level.
+         */
```
vmassol, line 85: Missing the javadoc for the parameter, that should trigger checkstyle normally

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
vmassol: I've now read more and this does 2 things
- replaces the default ErrorListener which logs to System.err
- swallows all logs for severities error and below, and only logs fatal errors

vmassol: > and are either handled in the {@link #extractXML} method

I don't think that's the case since the default error listener captures all exceptions and logs them, so they're never handled by extractXML AFAIK. Maybe I misunderstood what you meant?

vmassol:
@ClemensRobbenhaar We need a unit test for this and it's very simple to write. For the log capture you just need to use https://dev.xwiki.org/xwiki/bin/view/Community/Testing/JavaUnitTesting/#HCapturingandAssertingLogs

Let me know if you need my help, I'd be happy to help.

vmassol:
Here's a simple test that will output "length limit reached" and that you can modify to capture the log as mentioned above and assert it:

```
    @Test
    void extractXML() throws Exception
    {
        DOMImplementationLS lsImpl =
            (DOMImplementationLS) DOMImplementationRegistry.newInstance().getDOMImplementation("LS 3.0");
        LSInput input = lsImpl.createLSInput();
        input.setStringData("<root>hello world</root>");
        Document document = XMLUtils.parse(input);

        assertEquals("<root>llo</root>", XMLUtils.extractXML(document, 2, 4));
    }
```

vmassol:
And to get a fatal error, just change the input XML to:

```
 input.setStringData("<root>hello world</root><foo/>");
```

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
ClemensRobbenhaar: The default error listener captures all exceptions, but _does not handle_ them.

Instead the code inside the `Transformer` looks like this:

        } catch (RuntimeException e) {
            if (_errorListener != null) postErrorToListener(e.getMessage());
            throw new TransformerException(e);
        }

so any errors thrown by the transformation process still have to be handled by the calling side (just wrapped into a `TransformerException`) and are not swallowed by the error listener.

The calling side on xwiki-commons then seem to have the information to deal with these (wrapped) exceptions:

        } catch (Throwable t) {
            if (handler != null && handler.isFinished()) {
                return handler.getResult();
            } else {
                throw new RuntimeException("Failed to extract XML", t);
            }
        }

I am not sure if this is the right way to handle the exceptions but this is how the code is, independent on what I did in the pull request. Seemingly the way to handle this is first to check if the ``handler`` is finished - in this case the exception has been used for flow control, i.e. to signal that the rest of the document does not need to be processed (there seem to be no other way to stop the processing). Otherwise it is rethrown to be dealt with by the calling side (which probably shows an error message in the UI, but as this method can be called from many places I am not sure what it does).

In no case an exception is swallowed by the error listener. It is swallowed by the ``extractXML`` when that method guesses that the actual job to extract a text snippet has been successfully achieved.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
ClemensRobbenhaar: I can trigger checkstyle intentionally when building with `-Pquality`  and adding different kind of style violations _except in the javadoc_ for the inner class. It seems not to give a damn about it as it is private, and private constructs usually do not show up in the javadoc.

I will add a `@param exception the exception to be logged` to the javadoc, but I do not know anything more specific to say about the parameter. That documentation is a bit of a waste of space IMHO.

vmassol: Indeed the class is private, that's why. No need to do anything that doesn't bring value!

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -63,6 +64,79 @@
...
+        @Override
+        public void warning(TransformerException exception) throws TransformerException
+        {
+            LOGGER.debug("Warning [{}] from xml transformer", getRootMessage(exception));
+            if (LOGGER.isTraceEnabled()) {
+                LOGGER.trace(STACK_TRACE_NOTE, exception);
+            }
+        }
+
+        private String getRootMessage(Throwable t)
```
vmassol, line 128: No need for this. We are using ExceptionUtils already. It's called `getRootCauseMessage()`.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/XMLUtilsTest.java
```diff
@@ -279,4 +294,82 @@ void serializeNode()
...
+
+    @Test
+    void extractXMLWithTooLongTextNode() throws Exception
+    {
+        LSInput input = inputForExtractXML("<root>hello world</root>");
+        Document document = XMLUtils.parse(input);
+
+        assertEquals("<root>llo</root>", XMLUtils.extractXML(document, 2, 4));
+        assertEquals(Level.DEBUG, logCapture.getLogEvent(0).getLevel());
+        assertEquals(String.format("Error [%s] from xml transformer", "Length limit reached"),
```
vmassol, line 343: Personally I always assert the messsge we get, I feel it's more readable than recomputing it and it allows to more easily discover formatting problems. WDYT?

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
ClemensRobbenhaar: Ah, I did not know that. Updated in next commit.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/XMLUtilsTest.java
ClemensRobbenhaar: I first felt it might be more readable when separating the "inner" and "outer" message, but on second thought I agree with you.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -63,6 +65,68 @@
...
+         * These errors happen e.g. if a text node exceeds the expected
+         * maximal length.
+         *
+         * @param exception the exception to be logged
+         */
+        @Override
+        public void error(TransformerException exception) throws TransformerException
+        {
+            LOGGER.debug("Error [{}] from xml transformer", ExceptionUtils.getRootCauseMessage(exception));
+            if (LOGGER.isTraceEnabled()) {
```
vmassol, line 109: this is not needed and should be removed as best practice (it's not going to be more costly for perf)

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/XMLUtilsTest.java
```diff
@@ -279,4 +294,83 @@ void serializeNode()
...
+    @Disabled("because of unidentified location where code writes directly to the console")
+    void extractXMLWithBrokenInput() throws Exception
+    {
+        LSInput input = inputForExtractXML("<root>hello <b>world</a></root><garbage/>");
+        Document document = XMLUtils.parse(input);
+
+        assertNull(document, "we should not parse broken XML sucessfully");
+        assertEquals("", XMLUtils.extractXML(document, 2, 40));
+
+        // XXX: we expect the corresponding error message to go through our error listener
```
vmassol, line 357: We use `TODO` and not `XXX` for stuff to look at/fix in the future

PR #149, opened 2021-09-24
WIP: HTML Encoded Line Feed character transformed in whitespace

ClemensRobbenhaar:
This also upgrades the HTMLCleaner to version 2.25.

I also added some more test cases where I am not sure if they document wanted or unwanted behavior, or if they should be removed as they add nothing of value.
Because I am interested about feedback of the tests this is marked as "WIP" yet.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
@@ -109,7 +109,7 @@ public final class HTMLUtils
             String result = super.escapeElementEntities(text);
 
             // "\r" characters are automatically transformed in &#xD; but we want to keep the original \r there.
-            return cleanAmpersandEscape(result).replaceAll("&#xD;", "\r");
+            return result.replaceAll("&#xD;", "\r");
         }
 
         @Override
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
@@ -241,9 +241,6 @@ public class DefaultHTMLCleaner implements HTMLCleaner
         // See TrimAttributeCleanerTransformation for more information.
         defaultProperties.setTrimAttributeValues(false);
 
-        // This flag should be set to true once https://sourceforge.net/p/htmlcleaner/bugs/221/ is fixed.
-        defaultProperties.setRecognizeUnicodeChars(false);
-
         param = configuration.getParameters().get(HTMLCleanerConfiguration.TRANSLATE_SPECIAL_ENTITIES);
         boolean translateSpecialEntities = (param != null) && Boolean.parseBoolean(param);
         defaultProperties.setTranslateSpecialEntities(translateSpecialEntities);
--- a/xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
@@ -104,7 +104,9 @@ public class DefaultHTMLCleanerTest
         assertHTML("<p>\"&amp;</p>", "<p>\"&</p>");
         assertHTML("<p><img src=\"http://host.com/a.gif?a=foo&amp;b=bar\" /></p>",
             "<img src=\"http://host.com/a.gif?a=foo&amp;b=bar\" />");
-        assertHTML("<p>&#xA;</p>", "<p>&#xA;</p>");
+        assertHTML("<p>\n</p>", "<p>&#xA;</p>");
+        // the following is not a "special" character, but any ole unicode character
+        assertHTML("<p>ÿ</p>", "<p>&#xFF;</p>");
 
         // Verify that double quotes are escaped in attribute values
         assertHTML("<p value=\"script:&quot;&quot;\"></p>", "<p value='script:\"\"'");
@@ -520,7 +522,7 @@ public class DefaultHTMLCleanerTest
     }
 
     @Test
-    void controlCharacters() throws Exception
+    void controlCharactersAreReplacedBySpaces() throws Exception
     {
         String htmlInput = "<p>\u0008</p>";
         Document document = this.cleaner.clean(new StringReader(htmlInput));
@@ -533,25 +535,27 @@ public class DefaultHTMLCleanerTest
         htmlInput = "<p>&#8;</p>";
         document = this.cleaner.clean(new StringReader(htmlInput));
 
-        // HtmlCleaner currently doesn't handle properly unicode characters: asking it to recognize them
-        // involves that all entities will be escaped during the parsing and that's not what we want. So we
-        // keep them encoded.
+        // HtmlCleaner now handles unicode character entities. These are decoded
+        // if possible, and if they are control characters, replaced by whitespace
         // See https://sourceforge.net/p/htmlcleaner/bugs/221/
         textContent =
             document.getElementsByTagName("p").item(0).getTextContent();
-        assertEquals("&#8;", textContent);
-        assertHTML("<p>&#8;</p>", "&#8;");
+        assertEquals(" ", textContent);
+        assertHTML(" ", "&#8;");
 
         htmlInput = "<p foo=\"&#8;\">content</p>";
         document = this.cleaner.clean(new StringReader(htmlInput));
 
-        // HtmlCleaner currently doesn't handle properly unicode characters: asking it to recognize them
-        // involves that all entities will be escaped during the parsing and that's not what we want. So we
-        // keep them encoded.
+        // in attributes we actually want to keep (encoded) control characters
+        // but HTMLCleaner scraps these characters if it finds them in attributes
         textContent =
             document.getElementsByTagName("p").item(0).getAttributes().getNamedItem("foo").getTextContent();
-        assertEquals("&#8;", textContent);
-        assertHTML("<p foo=\"&#8;\">content</p>", "<p foo=\"&#8;\">content</p>");
+        // wanted behavior:
+        // assertEquals("&#8;", textContent);
+        // actual behavior
+        assertEquals("", textContent);
+        // same when serializing
+        assertHTML("<p foo=\"\">content</p>", "<p foo=\"&#8;\">content</p>");
     }
 
     private void assertHTML(String expected, String actual)
@@ -600,7 +604,7 @@ public class DefaultHTMLCleanerTest
         List<? extends TagNode> divList = tagNode.getElementListByName("div", true);
         assertEquals(1, divList.size());
         assertEquals("&quot;", divList.get(0).getText().toString());
-        // This assert is failing: the attribute is deserialized to contain ".
+        // This assert was failing: the attribute was deserialized to contain ".
         assertEquals("&quot;", divList.get(0).getAttributeByName("foo"));
 
         DomSerializer domSerializer = new DomSerializer(cleanerProperties, false);
@@ -615,7 +619,8 @@ public class DefaultHTMLCleanerTest
         nodeList = document.getElementsByTagName("div");
         assertEquals(1, nodeList.getLength());
         assertEquals("&quot;", nodeList.item(0).getTextContent());
-        // We can never retrieve the expected value here since the encoded &amp; has been lost earlier.
+        // We retrieve a &quot; here since the encoded &amp; is decoded to '&' when extracting
+        // the "text content" of the attribute
         assertEquals("&quot;", nodeList.item(0).getAttributes().getNamedItem("foo").getTextContent());
 
         assertHTML("<div foo=\"&amp;quot;\">content</div>",
@@ -623,13 +628,11 @@ public class DefaultHTMLCleanerTest
     }
 
     @Test
-    @Disabled("See https://sourceforge.net/p/htmlcleaner/bugs/221/")
     public void parseWithUnicodeChars() throws Exception
     {
         CleanerProperties cleanerProperties = new CleanerProperties();
 
         cleanerProperties.setDeserializeEntities(true);
-        cleanerProperties.setRecognizeUnicodeChars(true);
         cleanerProperties.setTranslateSpecialEntities(false);
 
         HtmlCleaner cleaner = new HtmlCleaner(cleanerProperties);
@@ -661,18 +664,104 @@ public class DefaultHTMLCleanerTest
 
         assertEquals("baz&buz", nodeList.item(2).getAttributes().getNamedItem("foo").getTextContent());
 
-        // BUG: This is failing with baz&gt;buz
+        // This was failing with baz&gt;buz
         assertEquals("baz>buz", nodeList.item(1).getTextContent());
-        // BUG: This is failing with baz&amp;buz
+        // This was failing with baz&amp;buz
         assertEquals("baz&buz", nodeList.item(2).getTextContent());
     }
 
+    /**
+     * Test that line feed and carriage return are inserted in their unencoded form.
+     */
     @Test
+    void doNotRemoveLFandCRInTextNodes()
+    {
+        assertHTML("<p>\r</p>", "<p>&#xD;</p>");
+        assertHTML("<p>Test \r\n</p>", "<p>Test \r\n</p>");
+    }
+
+    /**
+     * Ensure that &amp; in text nodes is not decoded even if it looks like it is following a character entity.
+     */
+    @Test
+    void doNotDecodeAmpersandInTextNodes()
+    {
+        assertHTML("<p>&amp;#xA;</p>", "<p>&amp;#xA;</p>");
+        assertHTML("<p>&amp;#10;</p>", "<p>&amp;#10;</p>");
+    }
+
+    /**
+     * Ensures that &lt; in text nodes is not decoded even if it is encoded as character entity.
+     */
+    @Test
+    void replaceOpeningBraceCharacterEntityByNamedEntity()
+    {
+        assertHTML("<p>&lt;</p>", "<p>&#60;</p>");
+        // actually it is not even decoded in attributes
+        assertHTML("<p foo=\"&lt;\">text</p>", "<p foo=\"&#60;\">text</p>");
+        // same if hexadecimal
+        assertHTML("<div>&lt;</div>", "<div>&#x003C;</div>");
+    }
+
+    /**
+     * Ensure that &gt; in text nodes is not decoded even if it is encoded as character entity.
+     */
+    @Test
+    void replaceClosingBracesCharacterEntityByNamedEntity()
+    {
+        assertHTML("<p>&gt;</p>", "<p>&#62;</p>");
+        // it is also not decoded in attributes (where it must not be decoded)
+        assertHTML("<p foo=\"&gt;\">text</p>", "<p foo=\"&#62;\">text</p>");
+        // same if hexadecimal
+        assertHTML("<div>&gt;</div>", "<div>&#x003E;</div>");
+    }
+
+    /**
+     * Ensures that &gt; in text nodes is encoded even if it is not encoded in the original text
+     */
+    @Test
+    void replaceBrokenExplicitClosingBraceByNamedEntity()
+    {
+        assertHTML("<p>&gt;</p>", "<p>></p>");
+        // however if a closing brace is inside an attribute, that is illegal html, and the html cleaner
+        // decides to move it to the next text node, where it is encoded (which is a legitimate decision)
+        // if this changes to something different, but also acceptable, feel free to update the test
+        assertHTML("<p foo=\"abc\">def\"&gt;text</p>", "<p foo=\"abc>def\">text</p>");
+    }
+
+    /**
+     * Test how unencoded opening angled braces aka greater-than sign are handled.
+     */
+    @Test
+    void replaceBrokenExplicitOpeningBraceByNamedEntity()
+    {
+        // ensures that &lt; in text nodes is encoded even if it is not encoded in the original text
+        assertHTML("<p>&lt;</p>", "<p><</p>");
+        // also, if it is surrounded by whitespace
+        assertHTML("<p>abc &lt; def</p>", "<p>abc < def</p>");
+        // however it does not, if it is followed by text; instead the brace and the following text are scrapped
+        assertHTML("<p>abc</p>", "<p>abc<def</p>");
+        assertHTML("<p>abc</p>", "<p>abc<def; and here we can have an much text as we want</p>");
+        // if an opening brace is inside an attribute, that is marginally ok, but the html cleaner
+        // decides to scrap it and all the following content
+        // could do this:
+        // assertHTML("<p foo=\"abc&lt;def\">text</p>", "<p foo=\"abc<def\">text</p>");
+        // but instead does:
+        assertHTML("<p foo=\"abc\">text</p>", "<p foo=\"abc<def\">text</p>");
+        // if surrounded by whitespace, then instead this happens
+        // not sure if this is good or bad; if the behavior of the HTML cleaner changes, feel free to update
+        assertHTML("<p foo=\"abc\">&lt; def\"&gt;text</p>", "<p foo=\"abc < def\">text</p>");
+    }
+
+    @Test
+    @Disabled("we have no special support for textarea at the moment")
     public void followingEncodedEntitiesAreProperlyKept()
     {
         String content = "<p><textarea>&#123;&#123;velocity}}machin&#123;&#123;/velocity}}</textarea></p>";
         Document document = this.cleaner.clean(new StringReader(content));
         String textareaContent = document.getElementsByTagName("textarea").item(0).getTextContent();
+        // this now produces "{{velocity}}machin{{;/velocity}} instead,
+        // as html entities in text are decoded unless necessary (like &gt; aka &#60;)
         assertEquals("&#123;&#123;velocity}}machin&#123;&#123;/velocity}}", textareaContent);
 
         assertHTML("<p><textarea>&#123;&#123;velocity}}machin&#123;&#123;/velocity}}</textarea></p>", content);
--- a/xwiki-commons-core/xwiki-commons-xml/pom.xml
+++ b/xwiki-commons-core/xwiki-commons-xml/pom.xml
@@ -32,7 +32,7 @@
   <packaging>jar</packaging>
   <description>XWiki Commons - XML</description>
   <properties>
-    <xwiki.jacoco.instructionRatio>0.72</xwiki.jacoco.instructionRatio>
+    <xwiki.jacoco.instructionRatio>0.74</xwiki.jacoco.instructionRatio>
     <!-- There's a utility class with lots of features, allow it to have many dependencies;
          There's a SAX event listener, which requires complex code -->
     <checkstyle.suppressions.location>${basedir}/src/main/checkstyle/checkstyle-suppressions.xml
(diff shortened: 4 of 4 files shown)


xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -104,7 +104,9 @@ void specialCharacters()
         assertHTML("<p>\"&amp;</p>", "<p>\"&</p>");
         assertHTML("<p><img src=\"http://host.com/a.gif?a=foo&amp;b=bar\" /></p>",
             "<img src=\"http://host.com/a.gif?a=foo&amp;b=bar\" />");
-        assertHTML("<p>&#xA;</p>", "<p>&#xA;</p>");
+        assertHTML("<p>\n</p>", "<p>&#xA;</p>");
```
ClemensRobbenhaar, line 107: this behavior is new in the 2.25 version: line breaks are now inserted verbatim instead of encoded.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -533,25 +535,27 @@ void controlCharacters() throws Exception
...
-        // keep them encoded.
+        // HtmlCleaner now handles unicode character entities. These are decoded
+        // if possible, and if they are control characters, replaced by whitespace
         // See https://sourceforge.net/p/htmlcleaner/bugs/221/
         textContent =
             document.getElementsByTagName("p").item(0).getTextContent();
-        assertEquals("&#8;", textContent);
-        assertHTML("<p>&#8;</p>", "&#8;");
+        assertEquals(" ", textContent);
+        assertHTML(" ", "&#8;");
```
ClemensRobbenhaar, line 544: Also new: control characters are now replaced by white space instead of keeping them encoded.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -533,25 +535,27 @@ void controlCharacters() throws Exception
...
+        // in attributes we actually want to keep (encoded) control characters
+        // but HTMLCleaner scraps these characters if it finds them in attributes
         textContent =
             document.getElementsByTagName("p").item(0).getAttributes().getNamedItem("foo").getTextContent();
-        assertEquals("&#8;", textContent);
-        assertHTML("<p foo=\"&#8;\">content</p>", "<p foo=\"&#8;\">content</p>");
+        // wanted behavior:
+        // assertEquals("&#8;", textContent);
+        // actual behavior
+        assertEquals("", textContent);
```
ClemensRobbenhaar, line 556: control characters in attribute are even removed completely.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -661,18 +664,104 @@ public void parseWithUnicodeChars() throws Exception
...
+    {
+        assertHTML("<p>&amp;#xA;</p>", "<p>&amp;#xA;</p>");
+        assertHTML("<p>&amp;#10;</p>", "<p>&amp;#10;</p>");
+    }
+
+    /**
+     * Ensures that &lt; in text nodes is not decoded even if it is encoded as character entity.
+     */
+    @Test
+    void replaceOpeningBraceCharacterEntityByNamedEntity()
```
ClemensRobbenhaar, line 697: Here I started writing even more tests to figure out how the 2.25 version behaves. These can be removed as well if the are only of personal interest for me than documenting wanted behavior (and possibly adding burden to future commits who need to keep these tests running).

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -661,18 +664,104 @@ public void parseWithUnicodeChars() throws Exception
...
+        // assertHTML("<p foo=\"abc&lt;def\">text</p>", "<p foo=\"abc<def\">text</p>");
+        // but instead does:
+        assertHTML("<p foo=\"abc\">text</p>", "<p foo=\"abc<def\">text</p>");
+        // if surrounded by whitespace, then instead this happens
+        // not sure if this is good or bad; if the behavior of the HTML cleaner changes, feel free to update
+        assertHTML("<p foo=\"abc\">&lt; def\"&gt;text</p>", "<p foo=\"abc < def\">text</p>");
+    }
+
     @Test
+    @Disabled("we have no special support for textarea at the moment")
```
ClemensRobbenhaar, line 757: I have not been able to keep this test running (except with weird hacks like encoding `&#123;` to `&amp;#123;` in the input stream, which are bound to break something else.) However I think the actual behavior is acceptable, too.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
```diff
@@ -241,9 +241,6 @@ private CleanerProperties getDefaultCleanerProperties(HTMLCleanerConfiguration c
         // See TrimAttributeCleanerTransformation for more information.
         defaultProperties.setTrimAttributeValues(false);
 
-        // This flag should be set to true once https://sourceforge.net/p/htmlcleaner/bugs/221/ is fixed.
-        defaultProperties.setRecognizeUnicodeChars(false);
-
```
ClemensRobbenhaar, line 246: as `true` is the default value for `recognizeUnicodeChars`, i just scrapped the setter.
Or should I leave it in as an explicit setting with some comment why this is important?

surli: IMO I do think it's better that we put the option value explicitely.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
```diff
@@ -109,7 +109,7 @@ public String escapeElementEntities(String text)
             String result = super.escapeElementEntities(text);
 
             // "\r" characters are automatically transformed in &#xD; but we want to keep the original \r there.
-            return cleanAmpersandEscape(result).replaceAll("&#xD;", "\r");
+            return result.replaceAll("&#xD;", "\r");
```
surli, line 112: This method is only called for unicode characters, or it's also called for other html entities? If it's only for unicode characters your change makes sense, else I'm not so sure that we want to remove the call to `cleanAmpersandEscape`.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
surli: I'm not sure we want this behavior to change.

surli (changes requested):
I don't think the PR is good as it is. AFAICS the unicode entities are now serialized using their "evaluated" value instead of keeping the actual entity value. Now the problem might now be related to the serialization of the DOM provided by HtmlCleaner. That needs a bit of investigation.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
ClemensRobbenhaar: This is called for all text nodes that are part of the parsed, sanitized DOM. It does not matter if they contain unicode or not.

The `cleanAmpersandEscape` has been removed exactly because of XCOMMONS-2265 : it replaces e.g. `&amp;#xD;`  by `&#xD;` (which is then displayed by browsers as as `<CR>`-control character), making it impossible for users to display a verbatim `&#xD;` in the display of the browser.  (Or to use any other encoded entity representation like "The Character α is represented by '&amp;#x3B1;'".  To display this information in HTML, one has to use:

    <p>The Character α is represented by &apos;&amp;#x3B1;&apos;</p>

(where the `&apos;` to represent the `'` is not mandatory, as we are outside an element tag, but the `&amp;` before the `#x3B1;` is, or the users would read "The Character α is represented by α", which is not the intended information.)

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
ClemensRobbenhaar: I guess I understand why you do not want to change this, but this is hard to achieve, if a literal `\n` line break should also be preserved in the output (as it is tested in `doNotRemoveLFandCRInTextNodes`) and implicitly in various other places.
The problem is that the input string is first converted into a DOM, and for text nodes in the DOM there is only the character `\n` to represent both  `&#xA;` and `\n` (and also `&#10;`, etc). The information if the character has been represented by its unicode entity representation or its raw value has been lost at this point.

When the `setRecognizeUnicodeChars` is set to false, however, any `&#xA;` is replaced by `&amp;#xA;` (as is the same for all other character entities), but then that is no longer distinguishable from the user input `&amp;#xA;`, which is the problem of XCOMMONS-2265.  The only way I see around this is either to replace `\n' or `&#xA;` by some marker string not present in the input string and then replace that marker string in the output by the character it represents. However that seems a very ugly hack for me.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
ClemensRobbenhaar: Ok, agreed. I put it back with an additional comment in the next commit.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
surli: Well we actually do perform automatic transformation of entities, the first assert of this test shows that we don't keep `&nbsp;` and `&quot;` as they are for example. So maybe unicode characters should be transformed to, note that we currently use in our default properties: `defaultProperties.setDeserializeEntities(true);` so this behaviour actually makes sense. 
I guess we need to ensure it won't cause any trouble for the users. I'm not entirely sure on my side, maybe @mflorea would have some feedback on that?

tmortagne: I don't see any reason to keep escaped characters that don't actually need escaping. We don't really lose any information.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
tmortagne: Never too many fast unit tests IMO. Make upgrades a lot safer.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
tmortagne: Yeah, I feel it's safer too, better be explicit and explain why we use this value just in case.

ClemensRobbenhaar: Put back setting the value for `recognizeUnicodeChars` in commit with some comment in https://github.com/xwiki/xwiki-commons/pull/149/commits/5cf1562d2cdcd6bad542a333ee8e53ca76ee5505

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -533,25 +535,27 @@ void controlCharacters() throws Exception
...
-        // keep them encoded.
+        // in attributes we actually want to keep (encoded) control characters
+        // but HTMLCleaner scraps these characters if it finds them in attributes
         textContent =
             document.getElementsByTagName("p").item(0).getAttributes().getNamedItem("foo").getTextContent();
-        assertEquals("&#8;", textContent);
-        assertHTML("<p foo=\"&#8;\">content</p>", "<p foo=\"&#8;\">content</p>");
+        // wanted behavior:
+        // assertEquals("&#8;", textContent);
+        // actual behavior
```
vmassol, line 555: I'm not sure I understand this. Are we testing for a behavior we don't want? that doesn't look good. Tests are the spec.

I think a better approach is to add a test with an ignore annotation and a description.

WDYT?

vmassol: Also is this a change from before the upgrade?

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -661,18 +664,104 @@ public void parseWithUnicodeChars() throws Exception
 
         assertEquals("baz&buz", nodeList.item(2).getAttributes().getNamedItem("foo").getTextContent());
 
-        // BUG: This is failing with baz&gt;buz
+        // This was failing with baz&gt;buz
```
vmassol, line 667: Why keep the comment if it's passing now?

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
vmassol: Agreed with Thomas.

vmassol: what's missing maybe is to explain why we want the behavior maybe...

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -661,18 +664,104 @@ public void parseWithUnicodeChars() throws Exception
...
+        assertHTML("<p>abc</p>", "<p>abc<def</p>");
+        assertHTML("<p>abc</p>", "<p>abc<def; and here we can have an much text as we want</p>");
+        // if an opening brace is inside an attribute, that is marginally ok, but the html cleaner
+        // decides to scrap it and all the following content
+        // could do this:
+        // assertHTML("<p foo=\"abc&lt;def\">text</p>", "<p foo=\"abc<def\">text</p>");
+        // but instead does:
+        assertHTML("<p foo=\"abc\">text</p>", "<p foo=\"abc<def\">text</p>");
+        // if surrounded by whitespace, then instead this happens
+        // not sure if this is good or bad; if the behavior of the HTML cleaner changes, feel free to update
```
vmassol, line 752: IMO, we need to remove these type of comments and decide what we want

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
ClemensRobbenhaar: Yes, this is a change introduced by the PR, by setting `defaultProperties.setRecognizeUnicodeChars(true)` for the HTML cleaner configuration.

Before the change, if some characters have embedded as entities (i.e. here `&#8;`) they have been inserted as such in the HTML and then "shown" in the browser as their character value (i.e. here a backspace).
 
However now they are first replaced by their actual, well, character (in this case as a single-byte character corresponding to `\b`, i.e backspace), because they need no entity encoding/escaping (only `&`, `<`, `>`, `"` and `'` do need it.).
In a second step the HTML-Cleaner detects those characters which are control characters and replaces them by white space (in text nodes) or removes them (in attribute text). I guess that is because the HTMLCleaner thinks control characters are not legit in HTML text or attributes nodes.

 Thus the attribute passed in as `foo="&#8;"`  is reduced to `foo=""`. in the previous version it was kept as `foo="&#8;"`. I kept the old behavior as  "wanted behavior" in the code as I am uncertain if this is wanted or not.

Note that line breaks are not subject to removal in text nodes. This is tested below in "doNotRemoveLFandCRInTextNodes"

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
ClemensRobbenhaar: Maybe I wanted to keep a note why we actually test this here. Then, if it is confusing, I am happy to scrap the comment.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
ClemensRobbenhaar: Actually in the case of the `replaceBrokenExplicitOpeningBraceByNamedEntity` I only document what the HtmlCleaner does, indeed. The input is invalid, so the HTML Cleaner has some leeway to decide what to create as output, because there is no unique solution.

If you say that tests only documenting the existing behavior are not good, I am ok with scrapping that test.

For the other test methods I will update the PR with a few more comments in the javadoc explaining why the test.

ClemensRobbenhaar:
I have updated the tests a bit, especially scrapping the one where it is not clear what the wanted output is.

Regarding the removal of control characters (even if encoded), which is a change brought by setting the `setRecognizeUnicodeChars(true)`, I m not sure if we want that, though I cannot think of a use case having these characters in the HTML. Also switching on this feature after a bug in it has been fixed is kinda the reason for this PR.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -661,18 +664,87 @@ public void parseWithUnicodeChars() throws Exception
...
+        assertHTML("<p>&gt;</p>", "<p>></p>");
+        // however if a closing brace is inside an attribute, that is illegal html, and the html cleaner
+        // decides to move it to the next text node, where it is encoded (which is a legitimate decision)
+        // if this changes to something different, but also acceptable, feel free to update the test
+        assertHTML("<p foo=\"abc\">def\"&gt;text</p>", "<p foo=\"abc>def\">text</p>");
     }
 
     @Test
+    @Disabled("we have no special handling for curly braces in textareas at the moment")
     public void followingEncodedEntitiesAreProperlyKept()
```
ClemensRobbenhaar, line 741: As I stumbled over: [AppWithinMinutes/Content.xml#L81](https://github.com/xwiki/xwiki-platform/blob/7b552f38974ef8ccbdd57f29b06a9049bb600376/xwiki-platform-core/xwiki-platform-appwithinminutes/xwiki-platform-appwithinminutes-ui/src/main/resources/AppWithinMinutes/Content.xml#L81)

    ## Escape {{ in the rendered content to be sure that the HTML macro is not closed unintentionally
    {{html}}$tdoc.getRenderedContent($tdoc.content, $tdoc.syntax.toIdString()).replace('{{', '&amp;#123;&amp;#123;'){{/html}}

I guess this test is relevant and I should not have disabled it.

However to keep the current behavior of html-escaping ``{{`` I could only think of a hack now used in a similar way for the `.replaceAll("&#xD;", "\r")` already done for the line breaks.

tmortagne: I'm actually not sure how this test is relevant, actually. What is important is the fact that the Velocity is escaping `{{` so that the HTML macro is not closed, but once you have the HTML content you don't really care what the HTML cleaner is doing, i.e. it does not matter if the RawBlock contains `{{`.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
surli: I guess it's ok given the first part of the test.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
surli: > However now they are first replaced by their actual, well, character (in this case as a single-byte character corresponding to \b, i.e backspace), because they need no entity encoding/escaping (only &, <, >, " and ' do need it.).

So that behaviour is checked first on previous part of the test, and I agree with it.

> In a second step the HTML-Cleaner detects those characters which are control characters and replaces them by white space (in text nodes) or removes them (in attribute text). I guess that is because the HTMLCleaner thinks control characters are not legit in HTML text or attributes nodes.

That second behaviour is a bit less clear: it would be interesting to make the same test before the HtmlCleaner upgrade with `\u0008` value in the attribute since AFAICS it should behave the same.

tmortagne:
By the way @ClemensRobbenhaar should something be done about https://github.com/xwiki/xwiki-commons/blob/3e6199463d8b093457b3e9204e67be5eaa9bcbc6/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/filter/ControlCharactersFilter.java#L42 in your pull request ?

ClemensRobbenhaar:
> By the way @ClemensRobbenhaar should something be done about
> 
> https://github.com/xwiki/xwiki-commons/blob/3e6199463d8b093457b3e9204e67be5eaa9bcbc6/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/filter/ControlCharactersFilter.java#L42
> in your pull request ?

Oh, I did not notice that class. I will check in the next days if we can get rid of it, and what might be the consequences.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
mflorea: @tmortagne it might be relevant. Here's an use case:

* you write in wiki syntax and save:
  ```
  {{html}}one&#123;&#123;/html}}two{{/html}}
  ```
* the output is as expected; you then edit with the WYSIWYG editor and save without any change
* the editor will send this to the sever:
  ```
  <!--startmacro:html|-||-|one&#123;&#123;/html}}two--><!--stopmacro-->
  ```
* the server will have to **clean** this and then convert to wiki syntax
* if the output of the ``HTMLCleaner`` is
  ```
  <!--startmacro:html|-||-|one{{/html}}two--><!--stopmacro-->
  ```
* then what does the XWiki Syntax Renderer produce? If it's this then it's bad:
  ```
  {{html}}one{{/html}}two{{/html}}
  ```

If this is the case, then saving with the WYSIWYG editor will produces changes whenever the HTMLCleaner replaces a character entity with the actual character (like it happens with Unicode characters after this change). For instance, this:

```
{{html}}&#xFF;{{/html}}
```

will end up in wiki syntax as:

```
{{html}}ÿ{{/html}}
```

after saving with the WYSIWYG editor. The users won't notice this while viewing the page but they will notice it in the page history (in the diff). I don't know if this is bad (except for ``{{`` which should be kept escaped) but in any case this is something to be aware of. We could justify it by saying that the WYSIWYG editor produces the canonical form.

mflorea (changes requested):

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
tmortagne: > `{{html}}one{{/html}}two{{/html}}`

This is not HTMLCleaner's problem. Properly escaping the content of a macro when rendering an XDOM to xwiki/2.1 syntax is xwiki/2.1 syntax renderer job.

> If this is the case, then saving with the WYSIWYG editor will produces changes whenever the HTMLCleaner replaces a character entity with the actual character (like it happens with Unicode characters after this change). 

That's nothing new, it's already the case for the wiki syntax when you wrote it without using the canonical form, and we decided long ago this was not a problem.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -457,6 +456,7 @@ void cleanHTMLTagWithNamespace()
      * check, see <a href="https://github.com/cure53/DOMPurify/commit/e32ca248c0e9450fb182e52e978631cbd78f1123">commit
      * e32ca248c0 in DOMPurify</a>.
      */
+    // FIXME: Test not passing anymore, apparently the template is surviving now.
```
surli, line 459: @michitux I need your opinion on this one: apparently test is not passing anymore in HTML5CleanerTest: the template tag is still there in the output.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
surli: I updated the comment with latest info from @mflorea

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
michitux: Maybe the tag definition in HTMLCleaner has been changed to be less strict? Is this still relevant (I see that you've closed the PR)?

surli: Yes I closed that PR because I moved the work to another branch to allow building this on CI, see: https://github.com/xwiki/xwiki-commons/pull/799#discussion_r1507810696

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
surli: @mflorea so my understanding is that `HTMLCleaner` behaviour did changed in 2.25 and it is actually main reason why @ClemensRobbenhaar wanted to upgrade, it fixes the bad escape of `&` see: https://sourceforge.net/p/htmlcleaner/bugs/221/

PR #157, opened 2022-01-04
XCOMMONS-901: Support both XHTML 1.0 and XHTML 5 in HtmlCleaner

michitux:
* Add a new HTML_VERSION parameter to HTMLCleanerConfiguration.
* Pass the HTML version in DefaultHTMLCleaner to HTMLCleaner and set the document type accordingly.
* Add a tag transformation to transform `<tt>` to `<span class="monospace">`.
* Add a tag transformation to replace the font filter in HTML5 as the cleaner removes the font tags too early otherwise.
* Extend BodyFilter to allow more HTML5 tags directly in the body.
* Add constants for many HTML5 tags in HTMLConstants.
* Add a new unit test HTML5HTMLCleanerTest to test the behavior of HTMLCleaner with the HTML5 configuration.
* Modify DefaultHTMLCleanerTest to allow re-using most test methods in HTML5HTMLCleanerTest.

Due to https://sourceforge.net/p/htmlcleaner/bugs/228/ and https://sourceforge.net/p/htmlcleaner/bugs/229/ there are currently two failing tests that I have disabled for HTML 5 (with a message mentioning these issues). I hope that these two bugs are not blockers for using the new parameter in XWiki. As the default is still HTML4 / XHTML 1.0 this should not break anything.

I haven't marked the new constants in HTMLConstants as unstable as they are very straightforward (apart from possible typos).

Jira issue: https://jira.xwiki.org/browse/XCOMMONS-901

### DIFF
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
@@ -318,8 +318,8 @@ public class DefaultHTMLCleaner implements HTMLCleaner
     {
         String param = configuration.getParameters().get(HTMLCleanerConfiguration.HTML_VERSION);
         int htmlVersion = 4;
-        if (param != null) {
-            htmlVersion = Integer.parseInt(param);
+        if ("5".equals(param)) {
+            htmlVersion = 5;
         }
         return htmlVersion;
     }
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
@@ -162,10 +162,11 @@ public class DefaultHTMLCleaner implements HTMLCleaner
             //   result = new DomSerializer(cleanerProperties, false).createDOM(cleanedNode);
 
             if (getHTMLVersion(configuration) == 5) {
-                cleanedNode.setDocType(new DoctypeToken("HTML", null, null, null));
+                cleanedNode.setDocType(new DoctypeToken(HTMLConstants.TAG_HTML, null, null, null));
             } else {
-                cleanedNode.setDocType(new DoctypeToken("html", "PUBLIC", "-//W3C//DTD XHTML 1.0 Strict//EN",
-                    "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd"));
+                cleanedNode.setDocType(
+                    new DoctypeToken(HTMLConstants.TAG_HTML, "PUBLIC", "-//W3C//DTD XHTML 1.0 Strict//EN",
+                        "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd"));
             }
             result =
                 new XWikiDOMSerializer(cleanerProperties).createDOM(getAvailableDocumentBuilder(), cleanedNode);
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
@@ -35,6 +35,7 @@ import org.htmlcleaner.DoctypeToken;
 import org.htmlcleaner.HtmlCleaner;
 import org.htmlcleaner.TagNode;
 import org.htmlcleaner.TagTransformation;
+import org.htmlcleaner.TrimAttributeTagTransformation;
 import org.htmlcleaner.XWikiDOMSerializer;
 import org.w3c.dom.Document;
 import org.xwiki.component.annotation.Component;
@@ -160,8 +161,12 @@ public class DefaultHTMLCleaner implements HTMLCleaner
             // Replace by the following when fixed:
             //   result = new DomSerializer(cleanerProperties, false).createDOM(cleanedNode);
 
-            cleanedNode.setDocType(new DoctypeToken("html", "PUBLIC", "-//W3C//DTD XHTML 1.0 Strict//EN",
-                "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd"));
+            if (getHTMLVersion(configuration) == 5) {
+                cleanedNode.setDocType(new DoctypeToken("HTML", null, null, null));
+            } else {
+                cleanedNode.setDocType(new DoctypeToken("html", "PUBLIC", "-//W3C//DTD XHTML 1.0 Strict//EN",
+                    "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd"));
+            }
             result =
                 new XWikiDOMSerializer(cleanerProperties).createDOM(getAvailableDocumentBuilder(), cleanedNode);
         } catch (ParserConfigurationException ex) {
@@ -232,9 +237,7 @@ public class DefaultHTMLCleaner implements HTMLCleaner
         defaultProperties.setTransResCharsToNCR(useCharacterReferences);
 
         // By default, we are cleaning XHTML 1.0 code, not HTML 5.
-        // Note: Tests are broken if we don't set the version 4, meaning that supporting HTML5 requires some work.
-        // TODO: handle HTML5 correctly (see: https://jira.xwiki.org/browse/XCOMMONS-901)
-        defaultProperties.setHtmlVersion(4);
+        defaultProperties.setHtmlVersion(getHTMLVersion(configuration));
 
         // We trim values by default for all attributes but the input value attribute.
         // The only way to currently do that is to switch off this flag, and to create a dedicated TagTransformation.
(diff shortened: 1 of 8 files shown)


vmassol:
About:

> Add a new unit test HTML5HTMLCleanerTest to test the behavior of HTMLCleaner with the HTML5 configuration.
> Modify DefaultHTMLCleanerTest to allow re-using most test methods in HTML5HTMLCleanerTest.

You could check parameterized tests in JUnit5, see https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests and especially `@MethodSource`

They should allow you to reuse the same tests and execute them with different results.

michitux:
> You could check parameterized tests in JUnit5, see https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests and especially `@MethodSource`

I actually looked at parameterized tests. There also seems to be the options to have parameters at class level, but the point is that I need to

* (Temporarily) disable some tests for HTML5 due to bugs in HTMLFilter.
* Probably also add some more tests that are just for HTML5.
* Change the expected results for some (few) tests.

At the same time, I want to test that the HTML5 option is as backwards-compatible as possible and to achieve this I really want to execute as many tests as possible for both configuration options as otherwise I wouldn't have noticed most of these breakages. I do not see how this is possible without either adding parameters to every single method or adding the parameters to the class but then I would need to have some "if parameter is HTML5, then execute different asserts" or "if parameter is HTML 4, do not execute this test". Therefore, to me it seemed cleaner to use a subclass for the HTML5 test. If you prefer parameterized tests instead, I can also try my luck with that.

I have also just noticed https://github.com/xwiki/xwiki-commons/compare/feature-html5cleaner - should have checked the Jira issue that linked these commits... It has some changes I do not have in `HTMLUtils`, I will see if they are necessary.

michitux:
> I have also just noticed https://github.com/xwiki/xwiki-commons/compare/feature-html5cleaner - should have checked the Jira issue that linked these commits... It has some changes I do not have in `HTMLUtils`, I will see if they are necessary.

Okay, I checked these changes and they are not necessary. Somehow the assumption back then was that we cannot set a document type for HTML5 and therefore need to manually print it but this is not true, we can set the document type also for HTML5. The old branch also uses a different approach for the font tag, it extends the definition of valid tags for HTML5 such that the font tag is valid and not stripped while my approach uses a new tag transformation. The transformation is less nice to configure externally, but as the font tag is really not a valid HTML 5 tag I think this is the cleaner solution.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
```diff
@@ -283,6 +286,16 @@ private CleanerTransformations getDefaultCleanerTransformations(HTMLCleanerConfi
         tt.addAttributeTransformation(HTMLConstants.ATTRIBUTE_STYLE, "text-align:center");
         defaultTransformations.addTransformation(tt);
 
+        if (getHTMLVersion(configuration) == 5) {
+            // Font tags are removed before the filters are applied in HTML5, we thus need a transformation here.
+            defaultTransformations.addTransformation(new FontTagTransformation());
+
+            tt = new TrimAttributeTagTransformation(HTMLConstants.TAG_TT,
+                HTMLConstants.TAG_SPAN);
+            tt.addAttributeTransformation(HTMLConstants.ATTRIBUTE_CLASS, "${class} monospace");
```
surli, line 295: Monospace is some kind of standard CSS class? Isn't it a font-family style?

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
```diff
@@ -295,4 +308,19 @@ private CleanerTransformations getDefaultCleanerTransformations(HTMLCleanerConfi
...
+    /**
+     * @param configuration The configuration to parse.
+     * @return The HTML version specified in the configuration.
+     * @since 14.0RC1
+     */
+    private int getHTMLVersion(HTMLCleanerConfiguration configuration)
+    {
+        String param = configuration.getParameters().get(HTMLCleanerConfiguration.HTML_VERSION);
+        int htmlVersion = 4;
+        if (param != null) {
```
surli, line 321: would be better to call `!StringUtils.isBlank(htmlVersion)` no? Unless you're sure the string cannot be empty?

surli (approved):
I made few comments, but once those are answered I think it's ok to be merged.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
mflorea: ``monospace`` CSS class is already used when rendering ``##foo## `` wiki syntax to HTML5, so @michitux is simply being consistent.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/FontTagTransformation.java
```diff
@@ -0,0 +1,96 @@
...
+        FONT_SIZE_MAP.put("4", "1.2em");
+        FONT_SIZE_MAP.put("5", "1.4em");
+        FONT_SIZE_MAP.put("6", "1.6em");
+        FONT_SIZE_MAP.put("7", "1.8em");
+        FONT_SIZE_MAP.put("-3", "0.4em");
+        FONT_SIZE_MAP.put("-2", FONT_SIZE_MAP.get("1"));
+        FONT_SIZE_MAP.put("-1", FONT_SIZE_MAP.get("2"));
+        FONT_SIZE_MAP.put("+1", FONT_SIZE_MAP.get("4"));
+        FONT_SIZE_MAP.put("+2", FONT_SIZE_MAP.get("5"));
+        FONT_SIZE_MAP.put("+3", FONT_SIZE_MAP.get("6"));
```
mflorea, line 55: Too bad the ``FontFilter`` didn't have any comment explaining how this mapping was obtained... It's not obvious at all.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/FontTagTransformation.java
```diff
@@ -0,0 +1,96 @@
...
+        if (attributes.containsKey(HTMLConstants.ATTRIBUTE_FONTCOLOR)) {
+            builder.append(String.format("color:%s;", attributes.get(HTMLConstants.ATTRIBUTE_FONTCOLOR)));
+        }
+        if (attributes.containsKey(HTMLConstants.ATTRIBUTE_FONTFACE)) {
+            builder.append(String.format("font-family:%s;", attributes.get(HTMLConstants.ATTRIBUTE_FONTFACE)));
+        }
+        if (attributes.containsKey(HTMLConstants.ATTRIBUTE_FONTSIZE)) {
+            String fontSize = attributes.get(HTMLConstants.ATTRIBUTE_FONTSIZE);
+            String fontSizeCss = FONT_SIZE_MAP.get(fontSize);
+            fontSizeCss = (fontSizeCss != null) ? fontSizeCss : fontSize;
```
mflorea, line 81:
```suggestion
            String fontSizeCss = FONT_SIZE_MAP.getOrDefault(fontSize, fontSize);
```

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/filter/BodyFilter.java
```diff
@@ -55,11 +55,17 @@
      * "P | %heading; | %list; | %preformatted; | DL | DIV | NOSCRIPT |
      * BLOCKQUOTE | FORM | HR | TABLE | FIELDSET | ADDRESS">
      * }</pre>
+     *
+     * We also use this list for HTML5 where in theory everything is allowed, but we instead only allow flow content
+     * that is not also phrasing content except for {@code <ins>} and {@code <del>} that were already allowed in HTML 4.
```
mflorea, line 60: This doesn't have any effect when filtering HTML4? Like allowing ``<article>`` and ``<section>`` inside the body tag when cleaning HTML4.

mflorea (commented):
Looks good overall, with small comments.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
michitux: Good point, though I guess we could alternatively also just do

```java
        if ("5".equals(param)) {
            htmlVersion = 5;
        }
```

as the only supported versions are 4 and 5. However, this ignores wrong values silently, no idea if this is okay or not.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/filter/BodyFilter.java
michitux: This list is not about allowing additional tags, this list is about which tags should not be wrapped in a `<p>`-tag when they are found as direct children below `<body>`. There are two reasons why I'm confident this is not a problem for cleaning HTML4:

* Most of these tags including `<article>` and `<section>` are not present in `HTML4TagProvider` and thus removed by HTMLCleaner before this filter is even called.  The tags `<menu>` and `<details>` are already in `HTML4TagProvider`, but they definitely shouldn't be wrapped in a paragraph also in HTML4 I think. An exception is `<template>`, where I just noticed that I did not follow the rule to wrap tags that are also phrasing content, but this feels like another exception that is worth making as `<template>`-tags are not rendered and thus having a paragraph around them seems wrong. I have extended the comment to explain this exception.
* Wrapping the listed tags in a paragraph seems wrong, regardless if we are dealing with HTML4 or HTML5.

mflorea (approved):

PR #160, opened 2022-01-14
XCOMMONS-901: Support both XHTML 1.0 and XHTML 5 in HtmlCleaner (follow-up)

michitux:
* Cleanup some comments and introduce a helper to check if the configuration is HTML 5.
* Allow b, i, u and s tags in HTML 5 as they are not obsolete in HTML 5 (and i is in fact frequently used for icons).
* Add XWikiHTML5TagProvider to fix svg and style tag handling.
* Fix class fan-out complexity in DefaultHTMLCleaner by not using the CleanerTransformations class but only our subclass.

This is a follow-up on https://jira.xwiki.org/browse/XCOMMONS-901 to fix some remaining issues with HTML 5 compatibility. Per the [development practices](https://dev.xwiki.org/xwiki/bin/view/Community/DevelopmentPractices#HRule:Don27tcreateunnecessaryissues) I haven't opened a separate issue for it.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLCleanerConfiguration.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLCleanerConfiguration.java
@@ -56,7 +56,7 @@ public interface HTMLCleanerConfiguration
     String TRANSLATE_SPECIAL_ENTITIES = "translateSpecialEntities";
 
     /**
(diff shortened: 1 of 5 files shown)


tmortagne:
@michitux @LGTM suggests that HTML5HTMLCleanerTest#conversionsFromHTML is failing.

michitux:
Should be fixed now, sorry, it seems I forgot to test again after disabling the transformations for the no-longer-deprecated tags in HTML5.

PR #165, opened 2022-02-15
XCOMMONS-2373: Support HTML 5 in HTML diff

michitux:
* Enable HTML 5 support in HTMLCleaner.
* Add a test case for HTML 5 support.

Jira issue: https://jira.xwiki.org/browse/XCOMMONS-2373


michitux:
Note that the build failure is due to https://jira.xwiki.org/browse/XCOMMONS-2151.

mflorea (approved):

PR #170, opened 2022-05-05
XCOMMONS-2426: Provide a component for filtering safe HTML elements and attributes

michitux:
* Add an interface and three implementations for an HTML element sanitizer.
* Let the default implementation dispatch to the different implementations depending on a configuration.
* Add tests.

Jira issue: https://jira.xwiki.org/browse/XCOMMONS-2426

### DIFF
--- a/xwiki-commons-core/xwiki-commons-xml/pom.xml
+++ b/xwiki-commons-core/xwiki-commons-xml/pom.xml
@@ -32,7 +32,7 @@
   <packaging>jar</packaging>
   <description>XWiki Commons - XML</description>
   <properties>
-    <xwiki.jacoco.instructionRatio>0.72</xwiki.jacoco.instructionRatio>
+    <xwiki.jacoco.instructionRatio>0.82</xwiki.jacoco.instructionRatio>
     <!-- There's a utility class with lots of features, allow it to have many dependencies;
          There's a SAX event listener, which requires complex code -->
     <checkstyle.suppressions.location>${basedir}/src/main/checkstyle/checkstyle-suppressions.xml
@@ -51,6 +51,11 @@
     <dependency>
       <groupId>org.xwiki.commons</groupId>
       <artifactId>xwiki-commons-context</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.xwiki.commons</groupId>
+      <artifactId>xwiki-commons-configuration-api</artifactId>
       <version>${project.version}</version>
     </dependency>
     <dependency>
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLElementSanitizer.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLElementSanitizer.java
@@ -1,0 +1,135 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.xml.internal.html;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Provider;
+import javax.inject.Singleton;
+
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.slf4j.Logger;
+import org.xwiki.component.annotation.Component;
+import org.xwiki.component.manager.ComponentLookupException;
+import org.xwiki.component.manager.ComponentManager;
+import org.xwiki.component.phase.Initializable;
+import org.xwiki.component.phase.InitializationException;
+import org.xwiki.configuration.ConfigurationSource;
+import org.xwiki.context.Execution;
+import org.xwiki.context.ExecutionContext;
+import org.xwiki.stability.Unstable;
+import org.xwiki.xml.html.HTMLElementSanitizer;
+
+/**
+ * Default {@link HTMLElementSanitizer} that loads the implementation chosen by the configuration.
+ *
+ * @version $Id$
+ * @since 14.6RC1
+ */
+@Component
+@Singleton
+@Unstable
+public class DefaultHTMLElementSanitizer implements HTMLElementSanitizer, Initializable
+{
+    private static final String CONFIGURATION_KEY = "xml.htmlElementSanitizer";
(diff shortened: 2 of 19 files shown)


xwiki-commons-core/xwiki-commons-xml/pom.xml
```diff
@@ -32,7 +32,7 @@
   <packaging>jar</packaging>
   <description>XWiki Commons - XML</description>
   <properties>
-    <xwiki.jacoco.instructionRatio>0.72</xwiki.jacoco.instructionRatio>
+    <xwiki.jacoco.instructionRatio>0.8</xwiki.jacoco.instructionRatio>
```
manuelleduc, line 35: Always use two digits for the ratio, so `0.80`

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/SecureHTMLElementSanitizer.java
```diff
@@ -0,0 +1,165 @@
...
+     * Attributes safe for values like "javascript:".
+     */
+    private final Set<String> uriSafeAttributes;
+
+    /**
+     * Default constructor.
+     */
+    public SecureHTMLElementSanitizer()
+    {
+        this.htmlTags = new HashSet<>(
```
tmortagne, line 90: I have the feeling that we should make this setup customizable through configuration, it would make easier for someone to workaround the fact that we are missing a safe element (new one or simply a mistake) for example.

michitux: I've actually wondered about this too. What do you suggest for the configuration variables, a comma-separated list of (fixed) names or a regular expression?

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLElementSanitizer.java
```diff
@@ -0,0 +1,97 @@
...
+@Singleton
+@Unstable
+public class DefaultHTMLElementSanitizer implements HTMLElementSanitizer, Initializable
+{
+    private static final String CONFIGURATION_KEY = "xml.htmlElementSanitizer";
+
+    private HTMLElementSanitizer implementation;
+
+    @Inject
+    @Named("xwikiproperties")
```
tmortagne, line 55: This concept does not really exist at xwiki-commons level (this configuration source implementation only exist in xwiki-platform) but you can obtain pretty much the same result without making DefaultHTMLElementSanitizer unusable outside of an XWiki instance by using

```
    @Inject
    @Named("restricted")
    Provider<ConfigurationSource> configurationSourceProvider;
```

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/SecureHTMLElementSanitizer.java
tmortagne: I was thinking about simple lists (xwiki-properties already has its own syntaxes for lists, so you can directly ask for one, no need to parse a string) but indeed you have a good point with the regex expression, there might be use cases where it's needed.

michitux: I would propose to first implement something similar to what the upstream source does. There, the configured allowed tags and attributes replace the default ones. I would propose to instead add extra tags and attributes via the configuration. Similar to the upstream, I would also add options to:
* Specify additional URI-safe attributes (i.e., attributes where arbitrary content should be allowed).
* Specify additional attributes where data-URIs shall be allowed.
* Forbid tags.
* Forbid attributes.

Should we also add the option to restrict URIs to allowed URIs with the option to override the [default regular expression](https://github.com/cure53/DOMPurify/blob/0965e9951dbd5bf79fee922dda2fc59f36ade0e7/src/regexp.js#L9)? I've so far just implemented the variant that allows arbitrary protocols (but forbids scripts and data-URIs) as it seemed quite restrictive to me and I would bet that this is the first thing where users complain that they cannot use their custom URIs to open their application. Upstream has URI restrictions as default and provides the option to allow all protocols, we could also reverse the default.

michitux:
I've addressed all comments. For my taste this is still lacking a few more tests in particular with the new configuration options but feel free to have a look at the changes already.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/SecureHTMLElementSanitizer.java
```diff
@@ -0,0 +1,235 @@
...
+    static final Pattern IS_SCRIPT_OR_DATA = Pattern.compile("^(?:\\w+script|data):", Pattern.CASE_INSENSITIVE);
+
+    static final Pattern ATTR_WHITESPACE =
+        Pattern.compile("[\\u0000-\\u0020\\u00A0\\u1680\\u180E\\u2000-\\u2029\\u205F\\u3000]");
+
+    static final Pattern DATA_ATTR = Pattern.compile("^data-[\\-\\w.\\u00B7-\\uFFFF]");
+
+    static final Pattern ARIA_ATTR = Pattern.compile("^aria-[\\-\\w]+$");
+
+    static final Pattern IS_ALLOWED_URI = Pattern.compile("^(?:(?:f|ht)tps?|mailto|tel|callto|cid|xmpp):",
```
tmortagne, line 69: I'm wondering if this (at least it's String part shouldn't be moved to being the default value of HTMLElementSanitizerConfiguration#getAllowedUriRegexp).

tmortagne (approved):

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/SecureHTMLElementSanitizer.java
michitux: I would agree for the URL regular expression as it seems like something that's relatively easy to understand for admins and admins need to provide a full regular expression, anyways. I wouldn't agree for the other properties, in particular for the list of allowed tags and attributes which could be moved into the configuration instead of having extra allowed/denied configurations. I decided for having configuration options that are empty by default for the following reasons:

* To support new HTML tags and attributes that might be required for new features in XWiki, we might need to add new tags and attributes. Admins who override the list might not have them which might result in strange bugs.
* For security reasons, it can be necessary to remove attributes and tags in security fix releases. Admins who override the list then won't automatically get the security fix which won't be noticeable but will make the XWiki installation vulnerable.

tmortagne: My point was not about changing the behavior, just moving the logic to HTMLElementSanitizerConfiguration.

michitux:
I've just pushed a rebase on the current master version and also an extension of the code to also allow SVG and MathML attributes and implemented [XCOMMONS-1680](https://jira.xwiki.org/browse/XCOMMONS-1680) based on this. I'll probably just change whole pull request to be about XCOMMONS-1680 and then just close XCOMMONS-2426 as implemented by XCOMMONS-1680. Further, I've decided to drop the "permissive" sanitizer option as I don't think this is maintainable. With the added configuration options, admins should be able to allow whatever attribute or element they need (and if they can't, we can add further options). I would still like to have more tests but I hope the remaining parts are already quite okay. 

I'm wondering if the sanitizer filter should force the secure sanitizer or if not, if it shouldn't check namespaces when the insecure sanitizer is used. Alternatively, we could also move the namespace check into the sanitizer but this is really quite specific to `HTMLCleaner` I think as in other cases you might also already have existing namespace information.

PR #206, opened 2022-05-11
[Misc] Always publish local builds to Gradle Enterprise

alextu:


manuelleduc:
Hello @alextu,

I'm asking out of curiosity. In your experience, what is the added benefit of publishing all the build including the local ones?
My initial feeling is that having the results from the CI is enough and having more would just be 'noise'. So I'd like to know what we could gain :)

Thanks.

alextu:
Sorry for the late response, we usually enable "always publish" for getting significant data as well as debugging issues (failing tests across various builds,...). But it's your call :)

PR #265, opened 2022-08-31
XCOMMONS-2507: Add a priority concept in components

surli:
* Modify `ComponentDescriptor` to introduce both roleHintPriority and
    roleTypePriority
  * Introduce support for `javax.annotation.Priority` to indicate the roleTypePriority
  * Modify ComponentAnnotationLoader to simplify computation if a component should be overridden by checking the roleHintPriority
  * Add support for roleHintPriority and roleTypePriority in EmbeddableComponentManager
  * Add support for role type priority in default ObservationManager

See https://github.com/xwiki/xwiki-platform/pull/2145 for the platform side of things.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-collection/src/main/java/org/xwiki/collection/internal/PriorityEntries.java
+++ b/xwiki-commons-core/xwiki-commons-collection/src/main/java/org/xwiki/collection/internal/PriorityEntries.java
@@ -139,8 +139,11 @@ public class PriorityEntries<E extends Comparable<E>>
     /**
      * Performs the given action for each value in this map until all entries have been processed or the action throws
      * an exception.
+     * <p>
+     * In this process, no specific order is guaranteed. If you need a fully sorted version, you should use
+     * {@link #getSorted()}.
      * 
-     * @param action the action to be performed for each element
+     * @param action the action to be performed for each value
      */
     public void forEachEntry(Consumer<E> action)
     {
@@ -156,9 +159,15 @@ public class PriorityEntries<E extends Comparable<E>>
     /**
      * @param key the key to insert
      * @param value the value to insert
+     * @exception NullPointerException if the key or value is {@code null}
      */
     public void put(String key, E value)
     {
+        // Make sure the value is not null
+        if (value == null) {
+            throw new NullPointerException();
+        }
+
         this.lock.writeLock().lock();
 
         try {
--- a/xwiki-commons-core/xwiki-commons-collection/src/main/java/org/xwiki/collection/internal/PriorityEntries.java
+++ b/xwiki-commons-core/xwiki-commons-collection/src/main/java/org/xwiki/collection/internal/PriorityEntries.java
@@ -1,0 +1,191 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.collection.internal;
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+
+/**
+ * A concurrent storage for a Map where values are sorted both by comparison and time of insertion.
+ * 
+ * @param <E> the type of the values stored in the map
+ * @version $Id$
+ * @since 15.4RC1
+ */
+public class PriorityEntries<E extends Comparable<E>>
+{
+    protected final Map<String, E> map;
+
+    protected List<E> sorted;
+
+    protected ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
+
+    /**
+     * Create a new empty {@link PriorityEntries}.
+     */
+    public PriorityEntries()
+    {
+        this.map = new LinkedHashMap<>();
+    }
+
+    /**
+     * Create a new empty {@link PriorityEntries}.
+     * 
+     * @param initialCapacity the initial capacity
+     */
+    public PriorityEntries(int initialCapacity)
+    {
+        this.map = new LinkedHashMap<>(initialCapacity);
+    }
+
+    /**
+     * @return the lock
+     */
+    public ReentrantReadWriteLock getLock()
+    {
+        return this.lock;
+    }
+
+    /**
+     * Returns the value to which the specified key is mapped, or {@code null} if this map contains no mapping for the
+     * key.
+     * 
+     * @param key the key whose associated value is to be returned
+     * @return the value to which the specified key is mapped, or {@code null} if this there is no mapping for the key
+     */
+    public E get(String key)
+    {
+        this.lock.readLock().lock();
+
+        try {
+            return this.map.get(key);
+        } finally {
+            this.lock.readLock().unlock();
+        }
+    }
+
+    /**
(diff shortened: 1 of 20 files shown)


xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/main/java/org/xwiki/component/embed/EmbeddableComponentManager.java
```diff
@@ -231,33 +233,43 @@ public <T> List<T> getInstanceList(Type role) throws ComponentLookupException
...
-        // Add local components
-        if (entries != null) {
-            for (Map.Entry<String, ComponentEntry<?>> entry : entries.entrySet()) {
+        Map<String, ComponentEntry<?>> entries = new HashMap<>();
+        if (this.componentEntries.containsKey(roleType)) {
+            entries.putAll(this.componentEntries.get(roleType));
+            for (ComponentEntry<?> value : entries.values()) {
                 try {
-                    components.put(entry.getKey(), getComponentInstance((ComponentEntry<T>) entry.getValue()));
+                    getComponentInstance(value); // force loading the instance
```
surli, line 241: Not very happy about this, but not sure how to do better for now (see comment below)

xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/main/java/org/xwiki/component/embed/EmbeddableComponentManager.java
```diff
@@ -231,33 +233,43 @@ public <T> List<T> getInstanceList(Type role) throws ComponentLookupException
...
+                if (!entries.containsKey(roleHint)) {
+                    T instance = getParent().getInstance(roleType, roleHint);
+                    entries.put(roleHint, new ComponentEntry<>(componentDescriptor, instance));
                 }
             }
         }
 
-        return components;
+        // Return a map ordered by the role type priority
+        return entries.entrySet().stream()
```
surli, line 263: @tmortagne I'm not super happy with this solution (hence the draft PR) because it means to compute back the ordering at each call. I'm a bit afraid about the perf impact, now I'm not sure how to do it better knowing that even if we order the components in the current manager, we'll need to reorder here because of the other components retrieved from the parents. wdyt?

tmortagne: I'm not sure we really need to order the Map. Shouldn't the ordering be done only in getInstanceList ?

tmortagne:
ComponentManager interface javadoc should be updated to mention the ordering.

xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/main/java/org/xwiki/component/embed/EmbeddableComponentManager.java
surli: Well the advantage to do it there, is that we automatically have the list ordered since it follows the iteration order of the map. Now if we decide to only have the ordering on the list I'm not sure it would be a lot better in perf. The only advantage I can see is that the code would be a bit simpler. 
Note that I haven't ordered the `#getComponentDescriptorList` either.

tmortagne: Yeah, I guess it does not hurt.

>  Note that I haven't ordered the #getComponentDescriptorList either.

Would be more consistent to apply the ordering there too since the order is provider by the descriptor.

surli: > Would be more consistent to apply the ordering there too since the order is provided by the descriptor.

Yes I was thinking about doing it, but I wanted to be sure that this first implementation was ok. So you don't think we could have strong performance issue here?

tmortagne: Ordering a few components is not a big issue IMO compared to the whole process of gathering those components. Especially since getComponentList is generally not used in a CPU intensive way.

I'm more concerned about the impact of the role hint priority support which, in theory, needs to select the component with the higher priority among all the members of the ComponentManager hierarchy before (initializing it if it's not already the case and) returning it, which is a big difference with current behavior. Now in theory pretty much mean one or two more Map#get, so the impact should not be that huge.

surli: > I'm more concerned about the impact of the role hint priority support which, in theory, needs to select the component descriptor with the highest priority among all the members of the ComponentManager hierarchy before (initializing it if it's not already the case and) returning it, which is a big difference with current behavior. Now in theory, it pretty much mean one or two more Map#get, so the impact should not be that huge.

Not sure to follow you there, have looked at my changes on `ComponentAnnotationLoader` (https://github.com/xwiki/xwiki-commons/pull/265/files#diff-7d816372e1d34e954cccbbf7b7d54f21181f1c92efc8bcd4248ecec56d62fab6)? I basically kept same implementation but I just avoid one map to directly use the priority info from the descriptor. So in theory, there should not be any performance change in there.

michitux: In theory, it should be possible to store the component descriptors in a `ConcurrentSkipListMap` sorted by priority. Then you would only need to merge the already sorted lists when getting the components from the parent. However, I agree that most likely the sorting here won't have any performance impact.

xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/main/java/org/xwiki/component/embed/EmbeddableComponentManager.java
```diff
@@ -196,27 +412,45 @@ public <T> T getInstance(Type roleType, String roleHint) throws ComponentLookupE
...
         }
 
         return instance;
     }
 
+    private <T> T getInstance(ComponentEntry<T> componentEntry) throws ComponentLookupException
+    {
+        try {
+            return getComponentInstance(componentEntry);
+        } catch (Throwable e) {
```
manuelleduc, line 445: Isn't sonar complaining for that? https://rules.sonarsource.com/java/RSPEC-1181

tmortagne: Yeah, I moved code all over the place. I guess there is a good chance sonar too thinks it's new code.

xwiki-commons-core/xwiki-commons-collection/src/main/java/org/xwiki/collection/internal/PriorityEntries.java
```diff
@@ -0,0 +1,191 @@
...
+     * @param key key whose mapping is to be removed from the map
+     * @return the previous value associated with {@code key}, or {@code null} if there was no mapping for {@code key}.
+     */
+    public E remove(String key)
+    {
+        this.lock.writeLock().lock();
+
+        try {
+            E value = this.map.remove(key);
+            if (value != null) {
```
michitux, line 182: If this data structure doesn't support `null` values, it might be better to explicitly check this I think (I think the other methods should support `null` values).

michitux: Actually, I've seen that other code explicitly puts `null` values, so I believe this `remove`-implementation is incorrect.

xwiki-commons-core/xwiki-commons-collection/src/main/java/org/xwiki/collection/internal/PriorityEntries.java
```diff
@@ -0,0 +1,191 @@
...
+                this.lock.readLock().unlock();
+            }
+        }
+
+        return this.sorted;
+    }
+
+    /**
+     * Performs the given action for each value in this map until all entries have been processed or the action throws
+     * an exception.
```
michitux, line 141: It might be clearer to state in which order the entries are processed.

xwiki-commons-core/xwiki-commons-collection/src/main/java/org/xwiki/collection/internal/PriorityEntries.java
tmortagne: Indeed, I don't have null values in the use cases I'm using this tool with, so I did not bother to take care of this, but would be better to be more explicit in other methods.

PR #285, opened 2022-10-05
XCOMMONS-2534: Provide a method to escape content passed to wiki macro list parameters

manuelleduc:
Jira: https://jira.xwiki.org/browse/XWIKI-20183

### DIFF
--- a/xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/EscapeTool.java
+++ b/xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/EscapeTool.java
@@ -32,6 +32,7 @@ import org.apache.commons.codec.net.QuotedPrintableCodec;
 import org.apache.commons.text.StringEscapeUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.xwiki.stability.Unstable;
 import org.xwiki.xml.XMLUtils;
 
 /**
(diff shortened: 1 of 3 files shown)


xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/EscapeTool.java
```diff
@@ -273,4 +276,14 @@ public String url(Object string)
         }
         return encodedURL;
     }
+
+    /**
+     * @return an instance of escapetool specific for {@code xwiki/21}
+     * @since 14.9RC1
+     */
+    @Unstable
+    public XWiki21EscapeTool getXwiki21()
```
vmassol, line 285: typo: it's `XWiki`:)

manuelleduc: Not if we want to be able to write `$escapetool.xwiki21.parameterArray` in velocity right?

vmassol:
See my answer at https://forum.xwiki.org/t/provide-more-escaping-helpers/11136/2?u=vmassol

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/EscapeTool.java
vmassol: Indeed (https://velocity.apache.org/engine/2.0/user-guide.html#property-lookup-rules), `getxwiki21()` would work too but it's not better.

vmassol: Maybe just put a comment to explain it's not a type but that it's for being able to write `$escapetool.xwiki21.`  from Velocity.

PR #323, opened 2022-12-02
XCOMMONS-1120: JSONTool is not escaping solidus (/)

michitux:
* Add custom character escapes for JSONTool that include forward slash.

Jira issue: https://jira.xwiki.org/browse/XCOMMONS-1120

### DIFF
--- a/xwiki-commons-core/xwiki-commons-velocity/src/test/java/org/xwiki/velocity/tools/JSONToolTest.java
+++ b/xwiki-commons-core/xwiki-commons-velocity/src/test/java/org/xwiki/velocity/tools/JSONToolTest.java
@@ -170,6 +170,13 @@ class JSONToolTest
     }
 
     @Test
+    void serializeSpecialCharacters()
+    {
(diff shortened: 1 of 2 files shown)


xwiki-commons-core/xwiki-commons-velocity/src/test/java/org/xwiki/velocity/tools/JSONToolTest.java
```diff
@@ -169,6 +169,13 @@ void serializeBean()
         assertFalse(json.contains("\"transientProperty\":\"transient\""));
     }
 
+    @Test
+    void serializeSpecialCharacters()
```
vmassol, line 173: maybe be more explicit like:  `serializeWhenHTMLCharacters()`?

michitux: I've now named it `serializeForwardSlash` which should be quite explicit.

PR #349, opened 2023-01-09
XCOMMONS-2602: Allow to retrieve an unified diff block corresponding to a specific line change

surli:


xwiki-commons-core/xwiki-commons-diff/xwiki-commons-diff-display/src/main/java/org/xwiki/diff/display/UnifiedDiffDisplayer.java
```diff
@@ -140,4 +141,26 @@ default <E, F> List<UnifiedDiffBlock<E, F>> display(DiffResult<E> diffResult, Li
...
+     * @param previousLineNumber the index of the line to be retrieved in case of deletion or context change, in case of
+     *                           addition the value should be negative
+     * @param nextLineNumber the index of the line  to be retrieved in case of addition or context change, in case of
+     *                          deletion the value should be negative
+     * @param configuration the same configuration used to compute the original diff
+     * @return a single diff block whose size is based on the selected line and on the size of the context
+     * @since 15.0RC1
+     */
+    @Unstable
+    <E, F> UnifiedDiffBlock<E, F> getContextDiffBlock(List<UnifiedDiffBlock<E, F>>
```
mflorea, line 164: I'm not sure ``UnifiedDiffDisplayer`` is the right place for this method. Judging from the input parameters, this method is called on the output produced by the unified diff displayer, so I think it make more sense to have it in a separate component. It's not easy to find a good name, maybe something like ``UnifiedDiffBlockExtractor``.

mflorea:
Overall looks of but if there's no need for this in XS then maybe you can keep the code in CR for now and later move it to commons when the need arises.

surli:
Closing this PR: there's no real need for it in XS, and even in change request I finally went to another solution (see: https://github.com/xwiki-contrib/application-changerequest/pull/90)

PR #481, opened 2023-07-24
XCOMMONS-2576: ReplayJob is not properly setting the parent job status of its children jobs

surli:
JIRA: https://jira.xwiki.org/browse/XCOMMONS-2756

Provide a setter for the parent job status in AbstractJobStatus, so that it's possible in ReplayJob to properly set the children job status' parent, without needing to rely on the context.


xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-api/src/main/java/org/xwiki/job/AbstractJobStatus.java
```diff
@@ -464,6 +465,20 @@ public JobStatus getParentJobStatus()
         return this.parentJobStatus;
     }
 
+    /**
+     * Allow to set the parent of this job.
+     * This setter is provided for specific usecases when one cannot set it from the context (e.g. from a replay job).
+     * @param parentJobStatus the actual parent job status
+     * @since 15.7-rc-1
```
surli, line 472: Not sure if it's for 15.7RC1 or if it can be for 15.6.

PR #482, opened 2023-07-25
[Misc] Moving deprecated Job APIs to legacy

surli:


xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/test/resources/jobs/status/id1/id2/id3/&status/status.xml
```diff
@@ -19,7 +19,7 @@
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
 -->
 
-<org.xwiki.job.internal.DefaultJobStatus>
+<org.xwiki.job.DefaultJobStatus>
```
surli, line 22: I had to modify the serialization in the XML files of the test, since the class is not present anymore in this module. I don't think it would have an impact for real world status serialization if the old class is available through the legacy module.

xwiki-commons-core/xwiki-commons-legacy/xwiki-commons-legacy-extension/xwiki-commons-legacy-extension-api/src/main/java/org/xwiki/extension/job/history/internal/LegacyExtensionJobHistoryRecorder.java
```diff
@@ -0,0 +1,65 @@
...
+@Deprecated
+public class LegacyExtensionJobHistoryRecorder extends ExtensionJobHistoryRecorder
+{
+    @Override
+    public void onEvent(Event event, Object source, Object data)
+    {
+        boolean ignore = false;
+        if (event instanceof JobStartedEvent) {
+            Job job = (Job) source;
+            if (job.getStatus() instanceof org.xwiki.job.AbstractJobStatus
```
surli, line 50: I might have used an aspect for injecting this condition check, but it felt easier to do it like that.

xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/job/history/internal/ExtensionJobHistoryRecorder.java
```diff
@@ -46,7 +46,7 @@
 import org.xwiki.job.event.JobStartedEvent;
 import org.xwiki.job.event.status.JobStatus;
 import org.xwiki.job.event.status.QuestionAnsweredEvent;
-import org.xwiki.job.internal.AbstractJobStatus;
+import org.xwiki.job.AbstractJobStatus;
```
surli, line 49: This impacts a condition in `#onJobStarted` which is called from `#onEvent` hence the legacy class I added to support the condition with the old class.

xwiki-commons-core/xwiki-commons-legacy/xwiki-commons-legacy-extension/xwiki-commons-legacy-extension-api/src/main/java/org/xwiki/extension/job/history/internal/LegacyExtensionJobHistoryRecorder.java
```diff
@@ -0,0 +1,65 @@
...
+ * Legacy implementation of {@link ExtensionJobHistoryRecorder}.
+ *
+ * @version $Id$
+ * @since 15.7RC1
+ */
+@Component
+@Named(ExtensionJobHistoryRecorder.NAME)
+@Singleton
+@Deprecated
+public class LegacyExtensionJobHistoryRecorder extends ExtensionJobHistoryRecorder
```
tmortagne, line 42: Seems to me this class is actually useless.

PR #508, opened 2023-09-08
Allow reusing compiled Velocity

tmortagne:
The idea is to be able to compile and cache Velocity scripts, and then reuse that cached AST in various contexts.

* https://jira.xwiki.org/browse/XCOMMONS-800
* https://jira.xwiki.org/browse/XCOMMONS-2795

Unfortunately, this comes with a breaking change: the previous way to handle Velocity scripts was a bit hackish, and we are now a bit more Velocity standard and one thing that change because of that is that macros defined in the script have priority over included macros (it actually used to be the opposite which was making impossible to override a macro included with something like #include* in the script itself, which IMO was probably quite a WTF).

As a bonus, Velocity errors are going to mention the actual script identifier instead of the current namespace (currently, a Velocity error mentions the namespace shared by all the scripts executed in the same scope which are including each other and not the actual script currently executed).


xwiki-commons-core/xwiki-commons-classloader/xwiki-commons-classloader-api/src/main/java/org/xwiki/classloader/ExtendedURLClassLoader.java
```diff
@@ -33,6 +33,20 @@
...
+    /**
+     * See {@link URLClassLoader#URLClassLoader(URL[], ClassLoader, URLStreamHandlerFactory)}.
+     *
+     * @param  name class loader name; or {@code null} if not named
+     * @param urls the URLs from which to load classes and resources
+     * @param parent the parent class loader for delegation
+     * @param factory the URLStreamHandlerFactory to use when creating URLs
+     * @since 15.8RC1
+     */
+    public ExtendedURLClassLoader(String name, URL[] urls, ClassLoader parent, URLStreamHandlerFactory factory)
```
surli, line 45: Missing unstable?

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/XWikiVelocityContext.java
```diff
@@ -110,6 +113,17 @@ private void warnDeprecatedBinding(String binding, ForeachScope foreachScope)
         }
     }
 
+    @Override
+    public void setMacroLibraries(List<Template> macroLibraries)
+    {
+        // Not super clean but that's the only way to not end up with current macro libraries being broken by
+        // Templace#parse
+        // TODO: remove when https://issues.apache.org/jira/browse/VELOCITY-967 is fixed
+        if (macroLibraries == null || !macroLibraries.isEmpty()) {
```
surli, line 122: The condition is weird but I guess it's expected.

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/VelocityEngine.java
```diff
@@ -45,7 +46,18 @@ public interface VelocityEngine
...
-     * Renders the input string using the context into the output writer.
+     * Compile the passed script into a Velocity {@link Template}.
+     * 
+     * @param name the name of the template
+     * @param source the input string containing the VTL to be rendered
+     * @return the compiler {@link Template}
+     * @throws XWikiVelocityException in case of error
+     * @since 15.8RC1
+     */
+    Template compile(String name, Reader source) throws XWikiVelocityException;
```
surli, line 57: Missing unstable.

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/VelocityEngine.java
```diff
@@ -71,6 +83,18 @@ public interface VelocityEngine
...
+     * Evaluate the template using the context into the output writer.
+     *
+     * @param context the Velocity context to use in rendering the input string
+     * @param out the writer in which to render the output
+     * @param namespace the string to be used as the template name for log messages in case of error. Also used as
+     *            namespace for the macros. Empty string means global namespace.
+     * @param template the compiled template to execute
+     * @throws XWikiVelocityException in case of error
+     */
+    void evaluate(Context context, Writer out, String namespace, Template template) throws XWikiVelocityException;
```
surli, line 96: Missing since and unstable

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/XWikiVelocityContext.java
tmortagne: Yes, it's weird. Not super happy about that one, but we'll see if Velocity guys accept my suggestion.

PR #547, opened 2023-10-05
XCOMMONS-2813: A compiled Velocity template should not be associated to a specific VelocityEngine

tmortagne:
Issue: https://jira.xwiki.org/browse/XCOMMONS-2813

The point of this change is to reduce even move the dependency on the context to compile a Velocity script (it used to depend on the current skin). As a bonus, xwiki-commons now comes with a fully usable Velocity manager.


xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/internal/DefaultVelocityManager.java
```diff
@@ -0,0 +1,264 @@
...
+        }
+    }
+
+    /**
+     * @param runtimeInstance the actual Velocity instance to use as reference
+     * @param name the name of the Velocity script
+     * @param source the source of the Velocity script
+     * @return the compiled script
+     */
+    public static VelocityTemplate compile(String name, Reader source, RuntimeInstance runtimeInstance)
```
surli, line 180: Static method inside a component? that's a bit surprising.

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/VelocityEngine.java
```diff
@@ -109,7 +102,10 @@ void evaluate(Context context, Writer out, String namespace, VelocityTemplate te
      *             sense anymore
      */
     @Deprecated
-    void clearMacroNamespace(String namespace);
+    default void clearMacroNamespace(String namespace)
+    {
+        
+    }
```
surli, line 108: Any reason for not moving those to legacy?

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/ScriptVelocityContext.java
```diff
@@ -0,0 +1,128 @@
...
+import org.apache.velocity.VelocityContext;
+import org.apache.velocity.runtime.directive.Scope;
+
+/**
+ * Maintains the current ScriptContext in sync with any modification of the VelocityContext.
+ * 
+ * @version $Id$
+ * @since 15.9-rc-1
+ */
+public class ScriptVelocityContext extends XWikiVelocityContext
```
surli, line 37: Missing unstable on that one.

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/internal/InternalVelocityEngine.java
```diff
@@ -37,34 +35,30 @@
...
 /**
  * Default implementation of the Velocity service which initializes the Velocity system using configuration values
  * defined in the component's configuration. Note that the {@link #initialize} method has to be executed before any
  * other method can be called.
  *
  * @version $Id$
+ * @since 15.9-rc-1
  */
-@Component
+@Component(roles = InternalVelocityEngine.class)
```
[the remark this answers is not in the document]
tmortagne: Using VelocityEngine as a component role is deprecated now and where InternalVelocityEngine is used I want to be sure it's a InternalVelocityEngine.

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/internal/DefaultVelocityManager.java
tmortagne: > Static method inside a component? that's a bit surprising.

We have many of those, actually. I just needed to put it in a place where I can reuse it and I don't want to expose that as an API.

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/ScriptVelocityContext.java
surli: Alors it feels like you could have a unit test for that one.

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/VelocityEngine.java
tmortagne: Mostly lazyness to add aspectJ setup xwiki-commons-legacy-velocity and all the plumbing than goes with it.

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/internal/DefaultVelocityManager.java
```diff
@@ -0,0 +1,264 @@
...
+        this.reservedBindings.add("foreach");
+        // Evaluate directive
+        this.reservedBindings.add("evaluate");
+        // TryCatch directive
+        this.reservedBindings.add("exception");
+        this.reservedBindings.add("try");
+        // Default directive
+        this.reservedBindings.add("define");
+        // The name of the context variable used for the template-level scope
+        this.reservedBindings.add("template");
```
surli, line 123: Just to be sure: there will never be a need to be able to add other custom reserved bindings? Wondering if those could be customized from a configuration property. I guess there will never be a need: if we need to add a binding it's because of an evolution of the velocity language no?

tmortagne: Not that all that is nothing new. I just moved DefaultVelocityManager from platform (but without the macros.vm stuff).

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/internal/InternalVelocityEngine.java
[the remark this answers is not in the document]
surli: > Using VelocityEngine as a component role is deprecated now

So shouldn't the interface itself be deprecated?

xwiki-commons-core/xwiki-commons-legacy/xwiki-commons-legacy-velocity/src/main/java/org/xwiki/velocity/VelocityFactory.java
```diff
@@ -30,8 +30,10 @@
  * we need this in XWiki since we want to allow each Skin to provide its own global macros.
  *
  * @version $Id$
+ * @deprecated use {@link VelocityManager} instead
  */
 @Role
+@Deprecated(since = "15.9-rc-1")
```
michitux, line 36: Don't we use the `15.9RC1` version format here? At least that's what I see when I look for other similar code.

xwiki-commons-core/xwiki-commons-legacy/xwiki-commons-legacy-velocity/src/main/java/org/xwiki/velocity/internal/DefaultVelocityEngine.java
```diff
@@ -0,0 +1,177 @@
...
+ * Default implementation of the Velocity service which initializes the Velocity system using configuration values
+ * defined in the component's configuration. Note that the {@link #initialize} method has to be executed before any
+ * other method can be called.
+ *
+ * @version $Id$
+ * @deprecated use {@link InternalVelocityEngine} instead
+ */
+@Component
+@InstantiationStrategy(ComponentInstantiationStrategy.PER_LOOKUP)
+@Deprecated(since = "15.9-rc-1")
```
michitux, line 54: Same here regarding the version format.

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/internal/InternalVelocityEngine.java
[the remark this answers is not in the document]
surli: At the very least it should be documented in the interface IMO. I feel like to be completely clean we'd need a new interface without the `@Role` and deprecate the current one, but maybe it's too much work for little benefits.

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/VelocityEngine.java
surli: ok

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/internal/DefaultVelocityManager.java
surli: ok

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/ScriptVelocityContext.java
```diff
@@ -0,0 +1,128 @@
...
+import javax.script.ScriptContext;
+
+import org.apache.velocity.VelocityContext;
+import org.apache.velocity.runtime.directive.Scope;
+
+/**
+ * Maintains the current ScriptContext in sync with any modification of the VelocityContext.
+ * 
+ * @version $Id$
+ * @since 15.9-rc-1
```
michitux, line 35: Same here, shouldn't this be 15.9RC1?

michitux (commented):
Hard to review this one, but I noticed that you're writing 15.9-rc-1 in many places where it should be 15.9RC1 I think.

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/ScriptVelocityContext.java
tmortagne: Right, looks like I forgot to move the unit test from platform along with ScriptVelocityContext.

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/internal/InternalVelocityEngine.java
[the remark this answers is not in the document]
tmortagne: > So shouldn't the interface itself be deprecated?

No it should not since that's what you are supposed to manipulate. You should just not directly try to inject one (it was already the case in practice before, you would not get a fully initialized engine).

> I feel like to be completely clean we'd need a new interface without the @Role and deprecate the current one, but maybe it's too much work for little benefits.

This interface never been used as component in practice (it never really provided a fully initialized engine), so I don't think it's needed.

I will just add a note that it should not be injected.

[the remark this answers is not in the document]
surli: ok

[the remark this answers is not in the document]
surli: Actually: why not officially breaking it and remove the `@Role` annotation then?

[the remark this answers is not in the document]
tmortagne: Not a fan of any breakage I can technically avoid, that's all.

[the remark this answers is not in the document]
surli: in general me neither, but here it feels that it would be the cleanest option on the long run. And according to your comment it's probably harmless. To be checked with other committers maybe.

xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/ScriptVelocityContext.java
tmortagne: Could not find any direct unit test in platform for the feature covered by ScriptVelocityContext (there might be some in the various other script macros tests), so I added one.

PR #656, opened 2023-11-24
XCOMMONS-2874: DefaultCoreExtensionScanner#loadEnvironmentExtension does not set extension url

manuelleduc:
Jira: https://jira.xwiki.org/browse/XCOMMONS-1099


xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/repository/internal/core/DefaultCoreExtensionScanner.java
```diff
@@ -362,4 +364,13 @@ private void fromXED(Map<String, DefaultCoreExtension> extensions, Collection<UR
             }
         }
     }
+
+    private static URL getEnvironmentExtensionURL(URL xedURL) throws MalformedURLException
+    {
+        String separator = "/";
+        List<String> segments = Arrays.asList(xedURL.toString().split(separator));
```
tmortagne, line 371: Not sure you really need to go through parsing a list, remove the ending "/META-INF/extension.xed" from the string would probably do it too.

tmortagne (approved):

PR #760, opened 2024-01-31
XCOMMONS-2921: Add support for overwritting the configuration from the execution context

mflorea:
# Jira URL

https://jira.xwiki.org/browse/XCOMMONS-2921

# Changes

## Description

* Implement a new "executionContext" configuration source
* Add ConfigurationSource#removeProperty()
* Add a helper to execute some code with modified configuration.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-configuration/xwiki-commons-configuration-api/pom.xml
+++ b/xwiki-commons-core/xwiki-commons-configuration/xwiki-commons-configuration-api/pom.xml
@@ -32,7 +32,7 @@
   <packaging>jar</packaging>
   <description>Provides the infrastructure for components needing configuration data</description>
   <properties>
-    <xwiki.jacoco.instructionRatio>0.00</xwiki.jacoco.instructionRatio>
+    <xwiki.jacoco.instructionRatio>0.54</xwiki.jacoco.instructionRatio>
     <!-- Name to display by the Extension Manager -->
     <xwiki.extension.name>Configuration API</xwiki.extension.name>
     <!-- Category to display in the Extension Manager -->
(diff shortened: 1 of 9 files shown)


xwiki-commons-core/xwiki-commons-configuration/xwiki-commons-configuration-api/src/main/java/org/xwiki/configuration/TemporaryConfigurationExecutor.java
mflorea, line 1: I'm not happy about this one, but I don't have a beer option. I'm open to suggestions.

mflorea:
Check https://github.com/xwiki/xwiki-platform/pull/2834 for the usage.

xwiki-commons-core/xwiki-commons-configuration/xwiki-commons-configuration-api/src/main/java/org/xwiki/configuration/TemporaryConfigurationExecutor.java
tmortagne: I'm not sure what we need is actually a configuration specific thing. To me, the real need is to pass a callable which is executed between an ExecutionContext set/unset and that's useful for various other use cases than just the configuration. You could simply call the default ConfigutationSource in the callable you pass to the helper.

For example, it could simply be something like `ExecutionContext#call(Map<String, Object> properties, Callable<V> callable)`

xwiki-commons-core/xwiki-commons-configuration/xwiki-commons-configuration-api/pom.xml
```diff
@@ -44,10 +44,21 @@
       <artifactId>xwiki-commons-component-api</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.xwiki.commons</groupId>
+      <artifactId>xwiki-commons-context</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
mflorea, line 51: The alternative was to put the code in ``xwiki-commons-context`` and add there the dependency on ``xwiki-commons-configuration-api``, but @tmortagne noted that it's more probable for a module that requires the configuration API to also require the context API, than the other way.

xwiki-commons-core/xwiki-commons-configuration/xwiki-commons-configuration-api/src/main/java/org/xwiki/configuration/TemporaryConfigurationExecutor.java
tmortagne: Actually, that proposal does not work well with custom storage in the ExecutionContext, which is most probably the case of the ExecutionContext ConfigurationSource.

PR #799, opened 2024-02-29
XCOMMONS-2276: Upgrade to HtmlCleaner 2.29

surli:
# Jira URL

* https://jira.xwiki.org/browse/XCOMMONS-2276
* https://jira.xwiki.org/browse/XCOMMONS-2265

# Changes

## Description

* Original discussion in #149 

## Clarifications

*

# Executed Tests

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  *


xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
```diff
@@ -113,7 +113,7 @@ public String escapeElementEntities(String text)
             String result = super.escapeElementEntities(text);
 
             // "\r" characters are automatically transformed in &#xD; but we want to keep the original \r there.
-            return cleanAmpersandEscape(result).replaceAll("&#xD;", "\r");
+            return result.replaceAll("&#xD;", "\r");
```
surli, line 116: This has to be double checked, see discussion in https://github.com/xwiki/xwiki-commons/pull/149/files#r778652172

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -457,7 +456,9 @@ void cleanHTMLTagWithNamespace()
      * check, see <a href="https://github.com/cure53/DOMPurify/commit/e32ca248c0e9450fb182e52e978631cbd78f1123">commit
      * e32ca248c0 in DOMPurify</a>.
      */
+    // FIXME: Test not passing anymore, apparently the template is surviving now.
     @Test
+    @Disabled
```
tmortagne, line 461: Does it mean the PR is not ready yet ? It's not marked as such.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
```diff
@@ -262,8 +262,9 @@ private CleanerProperties getDefaultCleanerProperties(HTMLCleanerConfiguration c
         // See TrimAttributeCleanerTransformation for more information.
         defaultProperties.setTrimAttributeValues(false);
 
-        // This flag should be set to true once https://sourceforge.net/p/htmlcleaner/bugs/221/ is fixed.
-        defaultProperties.setRecognizeUnicodeChars(false);
+        // Replace all unicode character entities by their actual unicode character, if possible.
+        // This flag is set to true since https://sourceforge.net/p/htmlcleaner/bugs/221/ is fixed in 2.25.
```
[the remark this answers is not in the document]
surli: Right, I guess @ClemensRobbenhaar set that to explain in the PR why he performed that change back then. Cleaning it up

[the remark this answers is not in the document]
surli: (Well actually was clear with the diff that removed the old comment...)

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
surli: Yes I need @michitux input on this, I forgot to mention it here too.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/internal/html/DefaultHTMLCleaner.java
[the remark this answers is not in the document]
vmassol: As a general comment, I think it's important to have comments explaining why we're not doing something , and not only explaining why why we do something. And even sometimes to have a little bit of history. I'm not saying it's the case here and I agree that the way it's phrased is a bit weird.

BTW:

> // Replace all unicode character entities by their actual unicode character, if possible.

Is not a good comment to me as it doesn't explain why we're doing this. It just explains the api called which is not super useful.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
michitux: As I noted, I believe that removing `<template>` tags inside `<select>` could be security-relevant. Unfortunately, I don't have more information than what's inside the commit I've linked to. The [standard](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#technical_summary) is pretty clear that the only allowed content of `<select>` is the following:

> Zero or more [`<option>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option), [`<optgroup>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup) or [`<hr>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr) elements.

Therefore, it is not correct to allow `<template>` elements inside `<select>`.

surli: Ok, thanks so you confirm we should keep it. I'll check how to handle this and if we report a bug to htmlcleaner for what look like a possible regression.

michitux: It should be possible to add a workaround for this on our side.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/html/HTMLUtils.java
surli: I believe it's ok as it's related to the bad escape that was fixed in: https://sourceforge.net/p/htmlcleaner/bugs/221/

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -125,7 +125,9 @@ void specialCharacters()
         assertHTML("<p>\"&amp;</p>", "<p>\"&</p>");
         assertHTML("<p><img src=\"http://host.com/a.gif?a=foo&amp;b=bar\" /></p>",
             "<img src=\"http://host.com/a.gif?a=foo&amp;b=bar\" />");
-        assertHTML("<p>&#xA;</p>", "<p>&#xA;</p>");
+        assertHTML("<p>\n</p>", "<p>&#xA;</p>");
+        // the following is not a "special" character, but any ole unicode character
```
tmortagne, line 129: Is this commend missing something ? Did you mean "any ole unicode character is escaped" ?

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
```diff
@@ -125,7 +125,9 @@ void specialCharacters()
         assertHTML("<p>\"&amp;</p>", "<p>\"&</p>");
         assertHTML("<p><img src=\"http://host.com/a.gif?a=foo&amp;b=bar\" /></p>",
             "<img src=\"http://host.com/a.gif?a=foo&amp;b=bar\" />");
-        assertHTML("<p>&#xA;</p>", "<p>&#xA;</p>");
```
tmortagne, line 128: Why remove that test ?

tmortagne:
Great to see we are finally going to upgrade HtmlCleaner :)

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
surli: The comment was part of changes performed by @ClemensRobbenhaar in the original PR. Honestly I don't know what's an "ole unicode character" so not sure how to fix the comment...

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
surli: I think we have other assertions checking this, but I'll check.

surli: Actually the assertion has not been removed it's the line just after: `assertHTML("<p>\n</p>", "<p>&#xA;</p>");`

tmortagne: No, it's not the same thing IMO.

surli: Well the assertion checks that `<p>&#xA;</p>` is transformed to `<p>\n</p>`. Not sure what else we can do here?

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
tmortagne: I think "ole unicode character" is related to characters with the two dot on top of it.

xwiki-commons-core/xwiki-commons-xml/src/test/java/org/xwiki/xml/internal/html/DefaultHTMLCleanerTest.java
surli: Yeah I first did same mistake...

PR #884, opened 2024-04-05
XCOMMONS-2989: Escape characters after < in XML comments

michitux:
# Jira URL

https://jira.xwiki.org/browse/XCOMMONS-2989

# Changes

## Description

* Refactor the code to avoid too high complexity
* Refactor the test to be parameterized and add a few test cases

## Clarifications

This is needed to fix https://jira.xwiki.org/browse/XWIKI-21868 and seems like the simplest, backwards-compatible change. An alternative would be to completely change the escaping format which would be a lot more breaking.

# Screenshots & Video

# Executed Tests

Built `xwiki-commons-xml` with the `quality` profile. This is very unlikely to cause any regressions as the escaping format remains unchanged.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * stable-15.10.x

### DIFF
--- a/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
+++ b/xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
@@ -248,6 +248,8 @@ public final class XMLUtils
      *   <li>3) Escape &gt; or - at the start of the comment</li>
      *   <li>4) Escape { to prevent XWiki macro syntax</li>
      *   <li>5) Add {@code \} (unescaped as {@code ""}) at the end if the last char is {@code -}</li>
+     *   <li>6) Ensure that {@code <} is always followed by {@code \} to avoid matching as HTML tag, e.g., in
+     *   CKEditor</li>
      * </ul>
      *
      * @param content the XML comment content to escape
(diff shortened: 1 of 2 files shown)


michitux:
The main change is basically the following:

```diff
-             if (c == '\\' || c == '{' || (c == '-' && lastChar == '-')) {
+             if (c == '\\' || c == '{' || (c == '-' && lastChar == '-') || lastChar == '<') {
```

The rest are comments and refactoring as this exceeds boolean and cyclomatic complexity thresholds.

xwiki-commons-core/xwiki-commons-xml/src/main/java/org/xwiki/xml/XMLUtils.java
```diff
@@ -248,6 +248,8 @@ public static String extractXML(Node node, int start, int length)
      *   <li>3) Escape &gt; or - at the start of the comment</li>
      *   <li>4) Escape { to prevent XWiki macro syntax</li>
      *   <li>5) Add {@code \} (unescaped as {@code ""}) at the end if the last char is {@code -}</li>
+     *   <li>6) Ensure that {@code <} is always followed by {@code \} to avoid matching as HTML tag, e.g., in
```
vmassol, line 251: * I guess this works because the HTML spec says that after `<` you cannot have the `\` character (when specifying a tag)? A reference to the spec would be nice.
* And we can use `\` because `\{char}` is just an escape in HTML comments and is the same as `{char}`?

This is maybe obvious but it could be interesting to mention it as I personally didn't find it so obvious.

PR #1075, opened 2024-09-02
Move to Jakarta based application servers

tmortagne:
See https://github.com/xwiki/xwiki-rendering/pull/318 for the rendering side of things.
See https://github.com/xwiki/xwiki-platform/pull/3388 for the platform side of things.

# Jira URLs

* [XCOMMONS-2108: Upgrade to Bean Validation 3.0.2](https://jira.xwiki.org/browse/XCOMMONS-2108)
* [XCOMMONS-2109: Upgrade to Hibernate Validator 8.0.1](https://jira.xwiki.org/browse/XCOMMONS-2109)
* [XCOMMONS-2475: Use Expressly instead of Apache EL](https://jira.xwiki.org/browse/XCOMMONS-2475)
* [XCOMMONS-2777: Support both javax.inject and jakarta.inject annotations](https://jira.xwiki.org/browse/XCOMMONS-2777)
* [XCOMMONS-2797: Allow accessing a javax.inject.Provider as a jakarta.inject.Provider and the opposite](https://jira.xwiki.org/browse/XCOMMONS-2797)
* [XCOMMONS-2962: Provide a javax/jakarta bridge for Servlet APIs](https://jira.xwiki.org/browse/XCOMMONS-2962)
* [XCOMMONS-2963: Upgrade to Servlet 5.0](https://jira.xwiki.org/browse/XCOMMONS-2963)
* [XCOMMONS-2994: Upgrade to Websocket 2.1.1](https://jira.xwiki.org/browse/XCOMMONS-2994)
*  [XWIKI-13963: Stop relying on the application server by default to resolve relative redirect URLs](https://jira.xwiki.org/browse/XWIKI-13963)
* [XWIKI-22586: Check all redirect URLs](https://jira.xwiki.org/browse/XWIKI-22586)
* [XWIKI-21137: Impossible to install XWiki Debian tomcat packages on Debian 12+ and Ubuntu 24.04+](https://jira.xwiki.org/browse/XWIKI-21137)

# Changes

## Description

The idea is to do just enough for XWiki to work in jakarta based Application servers (Jetty 11+, Tomcat 10+).

> [!IMPORTANT]
> This PR might look like a big change already but it's actually very far from a 100% jakarta XWiki. What was done is really just the strict minimum to work in a jakarta application server and there will still be a lot of work after that to fully migrate to jakarta.
> See https://design.xwiki.org/xwiki/bin/view/Proposal/Jakartamigration for more details.

### Jakarta WebSocket 2.1.1

XWiki has been migrated to WebSocket 2.1.1 without any kind of bridge, so every WebSocket related APIs are broken. There was no obvious way to keep supporting WebSocket 1 but fortunately no use of those APIs were found in contrib extensions.

### Jakarta Validation 3.0.2

XWiki has been migrated to Validation 3.0.2 (and Hibernate Validator 8.0.1) without any kind of bridge. A Javax Validation API JAR remain in the XWiki WAR because it's referenced by Jersey, but XWiki itself only support Jakarta Validation. In practice, it just means that Javax Validation annotations used in a contrib extension are going to be ignored. It's easy for an extension to support both Javax and Jakarta XWiki versions by declaring both Java Javax and Jakarta Validation annotations.

#### Migration to jakarta

* A `javax.servlet` <-> `jakarta.servlet` has been introduced.
* No Servlet related API should be broken.
* Except that the `Filter` and `Servlet` implementations declared in `web.xml` have all been moved to jakarta without keeping a javax compatible version. The reason is that there is no point keeping supporting those if XWiki is only used in jakarta based application servers, keeping the same class name also reduce the number of changes in the `web.xml` (in a sense, breaking them improve retro compatibility).
* All the oldcore servlet related APIs are now deprecated
  * The most commons Servlet related API has been added to the Container API to reduce the need to directly access it
  * A new script oriented Container API has been introduced
* Move to XJetty to Jetty 12 EE10 (instead of EE8)
* XWiki is now doing its best by default (in Jetty and Tomcat only right now) to enable everything it can find to allow any character in URLs at init
* New Tomcat 10 based Debian packages

#### Behavior changes

* Redirect URLs are now checked in all contexts (injected by a servlet filter for all requests)
* By default, relative redirect URLs are now resolved by XWiki instead of the application server (reduce setups requirement)

### Jakarta Inject 2.0.1

It was probably not strictly required, but both javax and jakarta Inject APIs are now supported at the same time.

* it did not made any sense for GenericProvider to be public, so I deprecated it and introduced two new internal tools (one for javax and the other for jakarta) to take care of the default Provider implementation/

# TODO

* run/fix Environment Tests
* would be good to optimize a bit more the size of the xjetty package (i.e. add some more JAR exclusions)

# Executed Tests

Not much new tests, but every single UI test and a lot of unit tests are impacted by this change so just the fact that they are passing is a good test. I also tried to migrate as little as possible from `javax` to `jakarta` as a way to test the bridges.

# Expected merging strategy

* Prefers squash: Yes


manuelleduc:
> I also tried to migrate as little as possible from javax to jakarta as a way to test the bridges.

@tmortagne it this a temporary step to validate the bridge? i.e., do we want to move all `javax` to `jakarta` quickly?

tmortagne:
> @tmortagne it this a temporary step to validate the bridge? i.e., do we want to move all javax to jakarta quickly?

The way I see it, the idea would be to migrate existing code only after a few releases, and we are happy with how things are behaving. But new code should obviously try to use jakarta APIs as much as possible (I tried to provide jakarta versions of all the APIs and to deprecate javax based one, at least for stuff which have been migrated already like servlet and inject).

xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/main/java/org/xwiki/component/annotation/ComponentDescriptorFactory.java
```diff
@@ -26,15 +26,16 @@
 import java.util.ServiceLoader;
 
 import javax.annotation.Priority;
```
michitux, line 28: There is no replacement for `Priority`? Or you just decided not to migrate that yet?

xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/main/java/org/xwiki/component/annotation/DefaultComponentDependencyFactory.java
```diff
@@ -39,30 +39,43 @@ public class DefaultComponentDependencyFactory extends AbstractComponentDependen
...
-        DefaultComponentDependency dependency;
+        DefaultComponentDependency dependency = null;
 
-        Inject inject = field.getAnnotation(Inject.class);
-        if (inject != null) {
-            dependency = new DefaultComponentDependency();
+        if (field.getAnnotation(Inject.class) != null || field.getAnnotation(javax.inject.Inject.class) != null) {
+            dependency = createDependency(field);
+        } else {
+            dependency = null;
```
michitux, line 47: This else doesn't seem necessary as the variable is already initialized with `null`. Also, isn't this a behavior change?

xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/test/java/org/xwiki/component/JavaxProviderIntegrationTest.java
```diff
@@ -19,6 +19,9 @@
  */
 package org.xwiki.component;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
```
michitux, line 24: According to https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HImports, static imports should be last, not first, so the change here seems wrong.

xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/main/java/org/xwiki/environment/internal/ServletEnvironment.java
```diff
@@ -27,10 +27,12 @@
 import java.net.URL;
 
 import javax.inject.Singleton;
-import javax.servlet.ServletContext;
 
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.xwiki.component.annotation.Component;
+import org.xwiki.jakartabridge.servlet.JakartaServletBridge;
+
+import jakarta.servlet.ServletContext;
```
michitux, line 35: I wonder if we should add a special rule for `jakarta` imports, e.g., to put them after `javax`?

xwiki-commons-core/xwiki-commons-jakartabridge/xwiki-commons-jakartabridge-common/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+    <artifactId>xwiki-commons-jakartabridge</artifactId>
+    <groupId>org.xwiki.commons</groupId>
+    <version>17.0.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>xwiki-commons-jakartabridge-common</artifactId>
+  <name>XWiki Commons - Jakarta Bridge - common</name>
+  <packaging>jar</packaging>
+  <description>Various tools to help bridge jakarta/javax APIs</description>
+  <properties>
+    <xwiki.jacoco.instructionRatio>0.0</xwiki.jacoco.instructionRatio>
```
michitux, line 34: No tests?

xwiki-commons-core/xwiki-commons-jakartabridge/xwiki-commons-jakartabridge-common/src/main/java/org/xwiki/jakartabridge/internal/JavaxToJakartaCollection.java
```diff
@@ -0,0 +1,150 @@
...
+    @Override
+    public Object[] toArray()
+    {
+        return toArray(EMPTY_OBJECT_ARRAY);
+    }
+
+    @Override
+    public <T> T[] toArray(T[] a)
+    {
+        if (a.length == 0) {
```
michitux, line 90: This looks wrong to me. According to my understanding, a new array should be allocated when the passed array is too small. Also, doesn't this mean that `toArray()` always returns an empty array?

xwiki-commons-core/xwiki-commons-jakartabridge/xwiki-commons-jakartabridge-common/src/main/java/org/xwiki/jakartabridge/internal/JavaxToJakartaCollection.java
```diff
@@ -0,0 +1,150 @@
...
+    public boolean remove(Object o)
+    {
+        return getJakarta().remove(this.jakartaProvider.apply((X) o));
+    }
+
+    @Override
+    public boolean containsAll(Collection<?> c)
+    {
+        return getJakarta()
+            .remove(JakartaBridge.toJakarta((Collection<X>) c, this.javaxProvider, this.jakartaProvider));
```
michitux, line 122: Calling `remove` in `containsAll` seems surprising.

xwiki-commons-core/xwiki-commons-jakartabridge/xwiki-commons-jakartabridge-servlet/src/main/java/org/xwiki/jakartabridge/servlet/JakartaServletBridge.java
```diff
@@ -0,0 +1,1002 @@
...
+     * @param jakarta the jakarta version
+     * @return the javax version
+     */
+    public static Enumeration<javax.servlet.Servlet> toJavax(Enumeration<Servlet> jakarta)
+    {
+        if (jakarta == null) {
+            return null;
+        }
+
+        return Collections.enumeration(Collections.list(jakarta).stream().map(JakartaServletBridge::toJavax).toList());
```
michitux, line 880: Converting first to a list seems quite a waste, according to [this answer](https://stackoverflow.com/a/33245847) it should also be possible to directly create a stream based on an iterator. Otherwise, I'm wondering if it wouldn't be cleaner to use a regular for-loop to directly create a list with the correct elements.

xwiki-commons-core/xwiki-commons-jakartabridge/xwiki-commons-jakartabridge-servlet/src/main/java/org/xwiki/jakartabridge/servlet/JakartaServletBridge.java
```diff
@@ -0,0 +1,1002 @@
...
+     * @param jakarta the jakarta version
+     * @return the javax version
+     */
+    public static EnumSet<javax.servlet.DispatcherType> toJavax(EnumSet<DispatcherType> jakarta)
+    {
+        if (jakarta == null) {
+            return null;
+        }
+
+        return EnumSet.copyOf(jakarta.stream().map(JakartaServletBridge::toJavax).collect(Collectors.toList()));
```
michitux, line 944: Why don't you directly collect the result of the stream in an `EnumSet`?

```suggestion
        return jakarta.stream().map(JakartaServletBridge::toJavax).collect(Collectors.toCollection(() -> EnumSet.noneOf(javax.servlet.DispatcherType.class)));
```

xwiki-commons-core/xwiki-commons-jakartabridge/xwiki-commons-jakartabridge-servlet/src/main/java/org/xwiki/jakartabridge/servlet/JakartaServletBridge.java
```diff
@@ -0,0 +1,1002 @@
...
+     * @param javax the javax version
+     * @return the javax version
+     */
+    public static EnumSet<DispatcherType> toJakarta(EnumSet<javax.servlet.DispatcherType> javax)
+    {
+        if (javax == null) {
+            return null;
+        }
+
+        return EnumSet.copyOf(javax.stream().map(JakartaServletBridge::toJakarta).collect(Collectors.toList()));
```
michitux, line 957: Same here as above.

xwiki-commons-core/xwiki-commons-jakartabridge/xwiki-commons-jakartabridge-servlet/src/main/java/org/xwiki/jakartabridge/servlet/JakartaServletBridge.java
```diff
@@ -0,0 +1,1002 @@
...
+    //////////////////////////////////////////////////
+    // Unsupported
+
+    /**
+     * @param jakarta the jakarta version
+     * @return the javax version
+     */
+    public static javax.servlet.descriptor.JspConfigDescriptor toJavax(JspConfigDescriptor jakarta)
+    {
+        // TODO
```
michitux, line 969: You still plan to implement these methods?

xwiki-commons-core/xwiki-commons-jakartabridge/xwiki-commons-jakartabridge-servlet/src/main/java/org/xwiki/jakartabridge/servlet/internal/JakartaToJavaxAsyncContext.java
```diff
@@ -0,0 +1,126 @@
...
+    public void addListener(AsyncListener listener, ServletRequest servletRequest, ServletResponse servletResponse)
+    {
+        this.javax.addListener(JakartaServletBridge.toJavax(listener), JakartaServletBridge.toJavax(servletRequest),
+            JakartaServletBridge.toJavax(servletResponse));
+    }
+
+    @Override
+    public <T extends AsyncListener> T createListener(Class<T> clazz) throws ServletException
+    {
+        // TODO
```
michitux, line 111: Do you still plan to implement this?

xwiki-commons-core/xwiki-commons-jakartabridge/xwiki-commons-jakartabridge-servlet/src/main/java/org/xwiki/jakartabridge/servlet/internal/JakartaToJavaxHttpServletRequest.java
```diff
@@ -0,0 +1,265 @@
...
+            return JakartaServletBridge.toJakarta(this.javax.getPart(name));
+        } catch (javax.servlet.ServletException e) {
+            throw new ServletException(e.getMessage(), e);
+        }
+    }
+
+    @Override
+    public <T extends HttpUpgradeHandler> T upgrade(Class<T> handlerClass) throws IOException, ServletException
+    {
+        // TODO
```
michitux, line 262: You still plan to implement this?

michitux (changes requested):
There seem to be subtle bugs in some of the adapters and converters. It would be great if we could have automated tests that protect against such bugs similar to the tests that I wrote for listeners, see, e.g., https://github.com/xwiki/xwiki-rendering/blob/9ac37383fe66215642ffc9f01e456c294a9f422f/xwiki-rendering-api/src/test/java/org/xwiki/rendering/listener/chaining/AbstractChainingListenerTest.java#L191. There also appear to be quite some TODOs left, do you plan to implement the missing parts or at least document clearly what is unsupported?

xwiki-commons-core/xwiki-commons-websocket/pom.xml
```diff
@@ -38,33 +38,38 @@
   </properties>
   <dependencies>
     <dependency>
-      <groupId>${project.groupId}</groupId>
+      <groupId>org.xwiki.commons</groupId>
```
michitux, line 41: How is this change related to the move to jakarta?

xwiki-commons-core/xwiki-commons-legacy/xwiki-commons-legacy-component/xwiki-commons-legacy-component-default/src/main/java/org/xwiki/component/embed/GenericProvider.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Default provider used when the Component Manager needs to inject a {@link Provider} field but no custom Provider has
+ * been registered. The default behavior is simply to look up the Component role class when {@link Provider#get()} is
+ * called. This is useful for example when you wish to do "lazy injection".
+ *
+ * @param <T> the role type
+ * @version $Id$
+ * @since 3.3M2
+ * @deprecated
```
michitux, line 37: Isn't this missing a comment which class should be used instead?

xwiki-commons-core/xwiki-commons-websocket/pom.xml
michitux, line 41: How is this change related to the move to jakarta?

xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/main/java/org/xwiki/component/annotation/ComponentDescriptorFactory.java
tmortagne: I did not really migrate `javax.annotation` yet. It was not really necessary for jakarta application servers.

xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/main/java/org/xwiki/component/annotation/ComponentAnnotationLoader.java
```diff
@@ -363,7 +363,11 @@ public Set<Type> findComponentRoleTypes(Class<?> componentClass, Type[] paramete
                     types.add(interfaceType);
                 }
 
-                // Handle javax.inject.Provider
+                // Handle javax.inject.Provider (retro-compatibility since **feature-deploy-jakarta**)
```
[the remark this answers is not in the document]
tmortagne: There are plenty of since to resolve, I need to be sure in which merge it's going to be merged first...

xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/test/java/org/xwiki/component/JavaxProviderIntegrationTest.java
tmortagne: Indeed, not sure how it ended there.

xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/main/java/org/xwiki/environment/internal/ServletEnvironment.java
tmortagne: It would indeed probably make sense. It's easy to configure in Eclipse, not sure how it works for IDEA.

xwiki-commons-core/xwiki-commons-jakartabridge/xwiki-commons-jakartabridge-common/src/main/java/org/xwiki/jakartabridge/internal/JavaxToJakartaCollection.java
tmortagne: This code is indeed wrong. Fixing.

xwiki-commons-core/xwiki-commons-jakartabridge/xwiki-commons-jakartabridge-servlet/src/main/java/org/xwiki/jakartabridge/servlet/JakartaServletBridge.java
tmortagne: I was lazy, but what would actually make the most sense in terms of performance is creating an Enumeration wrapper which manipulate the other one in its implementation, like I did for the other collections in `org.xwiki.jakartabridge.internal`. I'm going to add one more helper for Enumeration.

xwiki-commons-core/xwiki-commons-jakartabridge/xwiki-commons-jakartabridge-servlet/src/main/java/org/xwiki/jakartabridge/servlet/JakartaServletBridge.java
tmortagne: Not immediately, it did not seem to worth the trouble given that we have not using those concepts in XWiki. But eventually, probably.

xwiki-commons-core/xwiki-commons-jakartabridge/xwiki-commons-jakartabridge-servlet/src/main/java/org/xwiki/jakartabridge/servlet/JakartaServletBridge.java
tmortagne: I just did not notice this way to use collect yet.

xwiki-commons-core/xwiki-commons-legacy/xwiki-commons-legacy-component/xwiki-commons-legacy-component-default/src/main/java/org/xwiki/component/embed/GenericProvider.java
tmortagne: This should have never been public in the first place, there are only internal alternatives. I forgot about that one in the pull request description, indeed.

xwiki-commons-core/xwiki-commons-websocket/pom.xml
tmortagne: I doubt it is needed. I guess I just modified those by reflex for consistency.

xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/main/java/org/xwiki/environment/internal/ServletEnvironment.java
michitux: Easy to configure in IDEA, too, I think we provide the configuration with the configuration we provide for IDEA.

tmortagne: Sent https://forum.xwiki.org/t/add-a-special-ordering-location-for-jakarta-imports/16228.

xwiki-commons-core/xwiki-commons-jakartabridge/xwiki-commons-jakartabridge-servlet/src/main/java/org/xwiki/jakartabridge/servlet/JakartaServletBridge.java
michitux: It just seems strange to me that some random methods are missing without any indication in any documentation. I find this dangerous as it could cause random failures at runtime. You're certain that those methods that are missing won't be needed by some code that we don't control? I haven't fully checked where exactly those wrappers are used. In any case, if there is any risk, we should make sure that we don't use these wrappers in XS itself.

tmortagne: > In any case, if there is any risk, we should make sure that we don't use these wrappers in XS itself.

Unless I don't understand what you mean here, I already answered that: "we are not using those concepts in XWiki". Each of those methods is a rabbit hole of new wrappers to add.

xwiki-commons-core/xwiki-commons-legacy/xwiki-commons-legacy-component/xwiki-commons-legacy-component-default/src/main/java/org/xwiki/component/embed/GenericProvider.java
michitux: I would have just expected at least some comment here for potential (even if unlikely) users of that class what they should do. However, I just checked, and it's probably not used anymore after your change so it seems okay.

michitux (approved):

PR #1131, opened 2024-10-18
XCOMMONS-2174: Get rid of JDOM 1 dependency

michitux:
# Jira URL

https://jira.xwiki.org/browse/XCOMMONS-2174

# Changes

## Description

* Rewrite XML output in HTMLUtils using JDOM 2.
* Make the helper class private to not expose JDOM 2 in the API.
* Remove the dependency.

## Clarifications

* The class that was extended in the previous version is final in JDOM 2. Further, the methods that we're overriding are now in another class. Therefore, we're extending that class instead. To not repeat the mistake of exposing a dependency's API, I've made the nested class private.
* This pull request depends on xwiki/xwiki-platform#3576 and needs a vote for the breakage.

# Screenshots & Video

No changes.

# Executed Tests

In `xwiki-commons` (root):
```
LANG=C.UTF-8 mvn clean install -Pdocker,legacy,integration-tests,snapshotModules,quality,distribution,flavor-integration-tests
```

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * None (breaking change)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -577,11 +577,6 @@
       </dependency>
       <dependency>
         <groupId>org.jdom</groupId>
-        <artifactId>jdom</artifactId>
-        <version>1.1.3</version>
-      </dependency>
-      <dependency>
-        <groupId>org.jdom</groupId>
         <artifactId>jdom2</artifactId>
         <version>2.0.6.1</version>
       </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -2105,20 +2105,21 @@
               </rules>
             </configuration>
           </execution>
-          <!-- Check that we use the right version of the old jdom artifact -->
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -575,11 +575,6 @@
         <artifactId>dom4j</artifactId>
         <version>2.1.4</version>
       </dependency>
-      <dependency>
```
tmortagne, line 578: Would probably be good to also add an entry to make it a banned dependency.

michitux: I've added an exclude rule, extending the existing rule.

tmortagne: I had forgotten about the now useless `enforce-jdom` rule, indeed.

tmortagne (approved):

PR #1160, opened 2024-11-07
Remove extension empty folder from local storage upon extension removal

amottier:
Remove extension version folder and extension folder if they are empty when removing an extension from local repository.

# Jira URL

https://jira.xwiki.org/browse/XCOMMONS-3197

# Changes

## Description

* Call `File#delete` on `File#getParentFile` of the extension descriptor File to delete the folder named after the extension version. `File#delete` return `false` if directory is not empty so operation should be safe.
* Apply the same strategy to delete the folder named using the extension id.

# Executed Tests

* Existing unit tests.
* Manual testing.

# Expected merging strategy

* Prefers squash: Yes
* Backport on branches: ?

### DIFF
--- a/xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/repository/internal/local/LocalExtensionStorage.java
+++ b/xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/repository/internal/local/LocalExtensionStorage.java
@@ -306,5 +306,12 @@ public class LocalExtensionStorage
         if (extensionFile != null) {
             extensionFile.getFile().delete();
         }
+
+        // Delete extension version folder if empty
+        File extensionVesionFolder = descriptorFile.getParentFile();
+        extensionVesionFolder.delete();
+
+        // Delete extension folder if empty
+        extensionVesionFolder.getParentFile().delete();
     }
 }


xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/repository/internal/local/LocalExtensionStorage.java
```diff
@@ -306,5 +306,12 @@ public void removeExtension(DefaultLocalExtension extension) throws IOException
         if (extensionFile != null) {
             extensionFile.getFile().delete();
         }
+
+        // Delete extension version folder if empty
+        File extensionVesionFolder = descriptorFile.getParentFile();
```
vmassol, line 311: BTW there's a typo in the variable name ;)

xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/repository/internal/local/LocalExtensionStorage.java
```diff
@@ -306,5 +306,12 @@ public void removeExtension(DefaultLocalExtension extension) throws IOException
...
+        extensionVesionFolder.delete();
+
+        // Delete extension folder if empty
+        extensionVesionFolder.getParentFile().delete();
     }
```
vmassol, line 316: We should also handle the case when `delete()` returns false and at the very least either retry and/or log a warning. The existing code is also lacking this.

vmassol: We should also probably use java.nio.Files, see https://sonarcloud.io/organizations/xwiki/rules?open=java%3AS4042&rule_key=java%3AS4042

tmortagne: > We should also handle the case when delete() returns false and at the very least either retry and/or log a warning.

If it returns false, it usually just means it was not empty (which is something expected in many cases).

vmassol: Then I'd say it's a bad logic to call delete() without checking if it's empty.

amottier:
Thanks for the feedback.
I'll rework the PR to use java.nio and properly handle the exception.
And of course fix the typo!

vmassol:
@amottier just wanted to mention that your PR is good as-is IMO. My comments are more optional and ideas to improveeven further not just your PR but also the surrounding code.

xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/repository/internal/local/LocalExtensionStorage.java
amottier: In case that might be useful I give a try to use `java.nio.file` (only for `removeExtension`): https://github.com/xwiki/xwiki-commons/compare/master...amottier:fix-delete-extension-folder-on-removal-nio

If pertinent I can close this PR and open a new one with my `fix-delete-extension-folder-on-removal-nio` branch.

vmassol: I think it's good to apply this PR and then open a second one to improve the nio stuff if you want. Thx, I'll comment on the nio commits.

tmortagne: > We should also handle the case when delete() returns false

Note that Sonar and not too happy about that ignored return and fail the build, but https://github.com/xwiki/xwiki-commons/pull/1167 should fix this.

PR #1167, opened 2024-11-12
Attempt of using `java.nio.file` for extension removal

amottier:
This is a work in progress related to #1160

I'm opening the PR to track comments and feedback.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/repository/internal/local/LocalExtensionStorage.java
+++ b/xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/repository/internal/local/LocalExtensionStorage.java
@@ -304,17 +304,23 @@ public class LocalExtensionStorage
                 String.format("Extension [%s] does not exist: descriptor file is null", extension.getId().getId()));
         }
 
+        // Get extension file descriptor path
+        Path extensionDescriptorFilePath = descriptorFile.toPath();
+
+        // Delete the extension descriptor file
         try {
-            Files.delete(descriptorFile.toPath());
+            Files.delete(extensionDescriptorFilePath);
         } catch (FileNotFoundException e) {
             LOGGER.warn(
-                "Couldn't delete the extension descriptor file [{}] when removing extension [{}], because it doesn't exist. Root error: [{}]",
+                "Couldn't delete the extension descriptor file [{}] when removing extension [{}], "
+                    + "because it doesn't exist. Root error: [{}]",
                 descriptorFile.getAbsolutePath(), extension.getId().getId(), ExceptionUtils.getRootCauseMessage(e));
         }
 
         DefaultLocalExtensionFile extensionFile = extension.getFile();
 
         if (extensionFile != null) {
+            // Delete the extension file
             try {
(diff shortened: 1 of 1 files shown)


xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/repository/internal/local/LocalExtensionStorage.java
```diff
@@ -296,15 +300,61 @@ public void removeExtension(DefaultLocalExtension extension) throws IOException
...
+            try {
+                Files.delete(extensionFile.getFile().toPath());
+            } catch (FileNotFoundException e) {
+                LOGGER.warn("Extension file [{}] was not found while removing [{}] extension",
+                    extensionFile.getAbsolutePath(), extension.getId().getId());
+            }
+        }
+
+        // Get extension file descriptor path
+        Path extensionDescriptorFilePath = descriptorFile.toPath();
```
tmortagne, line 327: I would move that line above since you are using the same path to delete it.

xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/repository/internal/local/LocalExtensionStorage.java
```diff
@@ -296,15 +300,61 @@ public void removeExtension(DefaultLocalExtension extension) throws IOException
...
+
+        // Get the path to the folder that store the version of the extension being removed
+        Path extensionVersionFolderPath = extensionDescriptorFilePath.getParent();
+        try {
+            // Delete the extension version folder
```
tmortagne, line 332: I would be more explicit in the comment on the fact that it will only do something if it's empty.

xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/repository/internal/local/LocalExtensionStorage.java
```diff
@@ -296,15 +300,61 @@ public void removeExtension(DefaultLocalExtension extension) throws IOException
...
+            Files.delete(extensionDescriptorFilePath);
+
+            // Try to delete the extension folder
+            deleteExtensionFolderIfEmpty(extensionVersionFolderPath.getParent());
+        } catch (DirectoryNotEmptyException e) {
+            LOGGER.warn("Extension version folder [{}] was not empty after removing the extension [{}]. Keeping it.",
+                extensionDescriptorFilePath, extension.getId().getId());
```
tmortagne, line 339: Batter try to never pass complex objects like Path with potentially a lot of dependencies (problem when serializing and potentially dangerous to expose to script manipulating a job log) as log parameters.

tmortagne:
I did some minor modifications for it to be mergeable since it should fix the build current failing in master.

amottier:
> I did some minor modifications for it to be mergeable since it should fix the build current failing in master. Hope you don't mind @amottier.

Thanks a lot for finalizing this pull request, I initiate it but find my self not having the time finalize it in the pas few weeks.

PR #1220, opened 2025-01-27
Various changes due to issues building/running on Windows.

jameswartell:
Using the file class to build paths results in the file separator changes, but then we compare the paths using regex's which assume the Unix style file separator. There was also some encoding issues presumably due to UTF-8 not always being the default encoding on Windows.

# Jira URL

No Jira

# Changes

## Description

before using a regex to look at a path on systems where "/" is not the File.separator replace the native with the Unix style file.separator. 

Set the encoding to UTF-8 on every write/read in FormatMojoTest. 

*

## Clarifications

I discovered these issues trying to build the publication workflow application. The XAR XML files wouldn't validate because though some were set as visible technical pages, they didn't match the pattern given. I think we're thinking of these patterns to match paths like URL paths rather File paths, but the trouble comes in when we use the File class. I think this was the easiest way to fix it.

Then tests were failing due to encoding issues. I did add some debugging code to the tests to help pinpoint exactly where the first difference occurs on failure, because it was difficult to figure out what was going on at first. 

There were also System.lineSeperator() differences in the test. I think my GitHub client may convert text files on download to the native (Windows), but this results in String differences on a compare when you read the expected value from a file, since no such conversion happens when we write the xml in memory.

*

# Executed Tests

The tests were already failing, and it's a bit of a challenge to create the problem artificially. It would be good if builds were sometimes tested on Windows to find these issues.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
 
 Doesn't matter that I can see.

### DIFF
--- a/xwiki-commons-tools/xwiki-commons-tool-xar/xwiki-commons-tool-xar-plugin/src/main/java/org/xwiki/tool/xar/AbstractVerifyMojo.java
+++ b/xwiki-commons-tools/xwiki-commons-tool-xar/xwiki-commons-tool-xar-plugin/src/main/java/org/xwiki/tool/xar/AbstractVerifyMojo.java
@@ -19,6 +19,13 @@
  */
 package org.xwiki.tool.xar;
 
+import static org.twdata.maven.mojoexecutor.MojoExecutor.configuration;
+import static org.twdata.maven.mojoexecutor.MojoExecutor.element;
+import static org.twdata.maven.mojoexecutor.MojoExecutor.executeMojo;
+import static org.twdata.maven.mojoexecutor.MojoExecutor.executionEnvironment;
+import static org.twdata.maven.mojoexecutor.MojoExecutor.goal;
+import static org.twdata.maven.mojoexecutor.MojoExecutor.name;
+
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
@@ -44,13 +51,6 @@ import org.codehaus.plexus.components.io.fileselectors.IncludeExcludeFileSelecto
 import org.codehaus.plexus.components.io.resources.PlexusIoFileResourceCollection;
 import org.codehaus.plexus.components.io.resources.PlexusIoResource;
 
-import static org.twdata.maven.mojoexecutor.MojoExecutor.configuration;
-import static org.twdata.maven.mojoexecutor.MojoExecutor.element;
-import static org.twdata.maven.mojoexecutor.MojoExecutor.executeMojo;
-import static org.twdata.maven.mojoexecutor.MojoExecutor.executionEnvironment;
-import static org.twdata.maven.mojoexecutor.MojoExecutor.goal;
-import static org.twdata.maven.mojoexecutor.MojoExecutor.name;
-
 /**
  * Common code for the Verify and Format Mojos.
  *
@@ -72,6 +72,8 @@ public abstract class AbstractVerifyMojo extends AbstractXARMojo
     private static final String EXTENSION = ".xml";
 
     private static final Pattern TRANSLATION_PATTERN = Pattern.compile("(.*)\\..*\\.xml");
+
+    private static final String UNIX_STYLE_FILE_SEPARATOR = "/";
 
     /**
      * If true then don't check if the packaging is XAR before running mojos.
@@ -150,8 +152,8 @@ public abstract class AbstractVerifyMojo extends AbstractXARMojo
 
     /**
      * Explicitly define a list of pages (it's a regex) that are technical pages but that should be visible (not
-     * hidden). For example, home pages of applications. These pages must have their default langue set to empty so
-     * that a search in a given language doesn't return them as they're not content.
+     * hidden). For example, home pages of applications. These pages must have their default langue set to empty so that
+     * a search in a given language doesn't return them as they're not content.
      *
      * @since 12.3RC1
      */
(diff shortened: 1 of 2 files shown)


tmortagne:
A lot of formatting changes, which makes it hard to see the real changes. Better keep PR clean from such unneeded (and invalid for some of them) changes.

xwiki-commons-tools/xwiki-commons-tool-xar/xwiki-commons-tool-xar-plugin/src/test/java/org/xwiki/tool/xar/FormatMojoTest.java
```diff
@@ -146,7 +150,28 @@ void formatSpecialContentFailingWithXercesFromJDK8() throws Exception
         writer.setVersion("1.1");
         writer.write(domdoc);
         writer.close();
-
-        assertEquals(expectedContent, baos.toString());
+        String actual = baos.toString(Charset.forName("UTF-8"));
```
tmortagne, line 153: I think what you were searching for is `java.nio.charset.StandardCharsets.UTF_8`.

xwiki-commons-tools/xwiki-commons-tool-xar/xwiki-commons-tool-xar-plugin/src/test/java/org/xwiki/tool/xar/FormatMojoTest.java
```diff
@@ -132,8 +131,13 @@ void defaultLanguageForDocumentWhenNoTranslationButFileWithSameNameInOtherSpace(
     void formatSpecialContentFailingWithXercesFromJDK8() throws Exception
     {
         SAXReader reader = getSAXReader();
+        reader.setEncoding("UTF-8");
```
tmortagne, line 134: It's not clear to be why this is needed, as the encoding to use is indicated in the XML file.

xwiki-commons-tools/xwiki-commons-tool-xar/xwiki-commons-tool-xar-plugin/src/test/java/org/xwiki/tool/xar/FormatMojoTest.java
```diff
@@ -132,8 +131,13 @@ void defaultLanguageForDocumentWhenNoTranslationButFileWithSameNameInOtherSpace(
...
         InputStream is = Thread.currentThread().getContextClassLoader().getResourceAsStream("XWikiSyntaxLinks.it.xml");
         String expectedContent = IOUtils.toString(is, "UTF-8");
+        // github clients may automatically convert the file to use windows line returns.
```
tmortagne, line 137: Hmm, might be better to deal with this kind of problems with https://github.com/xwiki/xwiki-commons/blob/master/.gitattributes.

xwiki-commons-tools/xwiki-commons-tool-xar/xwiki-commons-tool-xar-plugin/src/main/java/org/xwiki/tool/xar/AbstractVerifyMojo.java
```diff
@@ -19,6 +19,13 @@
  */
 package org.xwiki.tool.xar;
 
+import static org.twdata.maven.mojoexecutor.MojoExecutor.configuration;
```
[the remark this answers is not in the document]
jameswartell: The eclipse formatter moved it. I needed to add this to the end in eclipse. I'll see about adding that to the section on setting up formatting in Eclipse.

image

xwiki-commons-tools/xwiki-commons-tool-xar/xwiki-commons-tool-xar-plugin/src/test/java/org/xwiki/tool/xar/FormatMojoTest.java
```diff
@@ -79,7 +80,7 @@ void defaultLanguageForMatchingContentPage()
         mojo.contentPages = Arrays.asList(".*/Document\\.xml");
         mojo.initializePatterns();
 
-        File file = new File("Some/Space/Document.xml");
+        File file = new File("Some" + File.separator + "Space" + File.separator + "Document.xml");
```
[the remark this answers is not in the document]
jameswartell: Probably don't actually. I think the file class will accept the unix style file.separator but when you call a method that returns a string you'll get whatever the native one is. I'll take it out and rebuild and see. There were several flavors of errors I was getting, so I made a bunch of changes just trying to make them all go away. Once I hit something that worked I probably didn't back everything out.

xwiki-commons-tools/xwiki-commons-tool-xar/xwiki-commons-tool-xar-plugin/src/test/java/org/xwiki/tool/xar/FormatMojoTest.java
jameswartell: Oh I was definitely just setting it absolutely everywhere I could, because it was frustrating trying to figure out why characters were invalid. Felt like whack-a-mole. I'll remove this one and see if I'm still broke.

xwiki-commons-tools/xwiki-commons-tool-xar/xwiki-commons-tool-xar-plugin/src/test/java/org/xwiki/tool/xar/FormatMojoTest.java
jameswartell: I dug in on this. It seemed intuitively like that might be a change with larger impacts. 

I tried setting the line.seperator on the format object of the writer to the native to see that would be an easy way to get them to match, but it did not. It is because the comment nodes of the parsed version have new lines in them which don't match the windows format. 

So I looked up if the SAX parser is misbehaving. It is not. The XML spec says: "To simplify the tasks of [applications](https://www.w3.org/TR/xml11/#dt-app), the [XML processor](https://www.w3.org/TR/xml11/#dt-xml-proc) MUST behave as if it normalized all line breaks in external parsed entities (including the document entity) on input, before parsing, by translating all of the following to a single #xA character:"
https://www.w3.org/TR/xml11/#sec-line-ends

That convinced me. I do think simply making all xml conform to using Unix style line returns is the right thing to do. The parser is right to strip out carriage returns. I believe using the EOL=LF in the .gitattributes will get clients to forcibly normalize this both locally and when checking in.

jameswartell:
> A lot of formatting changes, which makes it hard to see the real changes. Better keep PR clean from such unneeded (and invalid for some of them) changes.

I set up eclipse with xwiki formatter or it is happening by default. I didn't manually do those at all. I don't care about them either.

xwiki-commons-tools/xwiki-commons-tool-xar/xwiki-commons-tool-xar-plugin/src/test/java/org/xwiki/tool/xar/FormatMojoTest.java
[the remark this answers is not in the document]
jameswartell: confirmed, and removed

jameswartell:
Build of the xwiki-workflow-publication-api is now failing for me with this error when I try to use this as my snapshot.

`[ERROR Failed to execute goal morg.xwiki.commons:xwiki-commons-tool-xar-plugin:17.1.0-SNAPSHOT:verify(default-verify) on project xwiki-workflow-publication-application: Unable to execute mojo: Unable to parse configuration of mojo com.mycila:license-maven-plugin:3.0:check for parameter licenseSets: Cannot find 'licenseSets' in class com.mycila.maven.plugin.license.LicenseCheckMojo -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal morg.xwiki.commons:xwiki-commons-tool-xar-plugin:17.1.0-SNAPSHOT:verify(default-verify) on project xwiki-workflow-publication-application: Unable to execute mojo`

xwiki-commons-tools/xwiki-commons-tool-xar/xwiki-commons-tool-xar-plugin/src/main/java/org/xwiki/tool/xar/AbstractVerifyMojo.java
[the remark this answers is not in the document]
tmortagne: You might want to take a look at the various Eclipse related configurations on https://github.com/xwiki/xwiki-commons/tree/master/xwiki-commons-tools/xwiki-commons-tool-verification-resources/src/main/resources.

PR #1234, opened 2025-02-12
XCOMMONS-3258: Introduce a cache for ServletEnvironment#getResource

michitux:
# Jira URL

https://jira.xwiki.org/browse/XCOMMONS-3258

# Changes

## Description

* Introduce a dynamically initialized cache to avoid issues with recursive component initialization calls.
* Add test cases to confirm that the cache works as expected and that cache initialization can do recursive calls for getting resources.

## Clarifications

* As the proposal at https://forum.xwiki.org/t/cache-servletcontext-getresource/16416 and the discussion yesterday didn't bring a real conclusion about introducing a "development" mode, the cache can currently be disabled with the regular `CacheControl` mechanism that we're already using for other caches. In my opinion, we should discuss improving the cache control mechanism to limit it to users with programming right as this seems to be the easiest solution without introducing any side effects.
* Without the dynamic initialization, an endless recursion happens because the cache manager loads the configuration from the servlet environment, calling `getResource`.

# Screenshots & Video

No UI changes.

# Executed Tests

* `bin/dumbbench --float -p 0.1 -i 100 -- wget -q -o NUL -O NUL --load-cookies cookies.txt "http://127.0.0.1:1700/xwiki/bin/view/NoSpace/NoPage"` - the performance seems to be back to what it was approximately in 15.10.16, 103ms vs. 104ms in 15.10.16.
* `LANG=C.UTF-8 mvn clean install -Pdocker,legacy,integration-tests,snapshotModules,quality,distribution,flavor-integration-tests,standalone` in `xwiki-commons`.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * stable-16.10.x (as this fixes an important performance regression)

### DIFF
--- a/xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/main/java/org/xwiki/environment/internal/ServletEnvironment.java
+++ b/xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/main/java/org/xwiki/environment/internal/ServletEnvironment.java
@@ -85,7 +85,7 @@ public class ServletEnvironment extends AbstractEnvironment
                     // important to only request it once this component has been initialized.
                     CacheManager cacheManager = this.componentManager.getInstance(CacheManager.class);
                     this.resourceURLCache = cacheManager.createNewCache(
-                        new LRUCacheConfiguration("environment.servlet.resourceURLCache", 1000));
+                        new LRUCacheConfiguration("environment.servlet.resourceURLCache", 10000));
                 }
             } catch (Exception e) {
                 this.logger.error("Failed to initialize resource URL cache.", e);
--- a/xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/main/java/org/xwiki/environment/internal/ServletEnvironment.java
+++ b/xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/main/java/org/xwiki/environment/internal/ServletEnvironment.java
@@ -25,13 +25,21 @@ import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
 
+import javax.inject.Inject;
 import javax.inject.Singleton;
 
 import jakarta.servlet.ServletContext;
 
 import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.xwiki.cache.Cache;
+import org.xwiki.cache.CacheControl;
+import org.xwiki.cache.CacheManager;
+import org.xwiki.cache.config.LRUCacheConfiguration;
 import org.xwiki.component.annotation.Component;
+import org.xwiki.component.manager.ComponentManager;
 import org.xwiki.jakartabridge.servlet.JakartaServletBridge;
 
 /**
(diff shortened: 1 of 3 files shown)


xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/main/java/org/xwiki/environment/internal/ServletEnvironment.java
```diff
@@ -54,6 +62,41 @@ public class ServletEnvironment extends AbstractEnvironment
...
+            try {
+                // Compare again inside the "lock" to avoid race conditions. Only after seeing the false value we can
+                // be sure that we also see the resourceURLCache variable write from the thread that wrote it -
+                // writing atomics with "set" guarantees sequential consistency.
+                if (this.resourceURLCache == null) {
+                    // The cache manager can't be injected directly as it depends on this component, so it is
+                    // important to only request it once this component has been initialized.
+                    CacheManager cacheManager = this.componentManager.getInstance(CacheManager.class);
+                    this.resourceURLCache = cacheManager.createNewCache(
+                        new LRUCacheConfiguration("environment.servlet.resourceURLCache", 1000));
```
surli, line 88: Isn't 1000 a bit low value by default?

xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/main/java/org/xwiki/environment/internal/ServletEnvironment.java
```diff
@@ -54,6 +62,41 @@ public class ServletEnvironment extends AbstractEnvironment
      */
     private javax.servlet.ServletContext javaxServletContext;
 
+    @Inject
+    private ComponentManager componentManager;
```
surli, line 66: I think it's ok but just to be sure: we agree it wouldn't make sense to use the contextual CM here? (`@Named("context")`)?

xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/main/java/org/xwiki/environment/internal/ServletEnvironment.java
```diff
@@ -54,6 +62,41 @@ public class ServletEnvironment extends AbstractEnvironment
...
+
+    @Inject
+    private CacheControl cacheControl;
+
+    private Cache<Optional<URL>> resourceURLCache;
+
+    private final AtomicBoolean cacheInitializing = new AtomicBoolean();
+
+    private Cache<Optional<URL>> getResourceURLCache()
```
tmortagne, line 75: In case you did not have it in mind, an alternative to this complex ordering game would be to put the cache right in a proxy of the ServletContext.

xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/main/java/org/xwiki/environment/internal/ServletEnvironment.java
tmortagne: I don't think it would make much sense to use the context since that componentManager is used way before any concept of context event exist.

xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/main/java/org/xwiki/environment/internal/ServletEnvironment.java
tmortagne: At least it feels like we could go up without it being noticeable much in terms of memory.

tmortagne (approved):

xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/test/java/org/xwiki/environment/internal/ServletEnvironmentTest.java
```diff
@@ -22,37 +22,71 @@
...
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.when;
 
 /**
  * Unit tests for {@link ServletEnvironment}.
  *
  * @version $Id$
  * @since 3.5M1
  */
+@SuppressWarnings({ "checkstyle:ClassFanOutComplexity", "checkstyle:MultipleStringLiterals" })
```
tmortagne, line 72: Hmm, I though checkstyle was not applied to test classes, but maybe not anymore.

tmortagne: Fine with me, I was just wondering.

xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/main/java/org/xwiki/environment/internal/ServletEnvironment.java
tmortagne: 10k sounds good

surli: :+1: was also thinking of 10K

xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/main/java/org/xwiki/environment/internal/ServletEnvironment.java
surli: All good then.

xwiki-commons-core/xwiki-commons-environment/xwiki-commons-environment-servlet/src/test/java/org/xwiki/environment/internal/ServletEnvironmentTest.java
surli: > Honestly, I would not be against enabling Checkstyle in the test classes, but we would need to ignore a lot of the existing classes I guess.

yep it will take time until we achieve that... ideally we should probably enable it by default and have a property to switch it off in all modules already existing until we fix them.

tmortagne: This can be a good Quality Day (next week) pick by the way ;)

PR #1243, opened 2025-02-19
updated CheckStyle config 

Anmol202005:
# Updated Checkstyle Configuration: ClassFanOutComplexity max value

## Summary
This PR updates the checkstyle configuration to increase the `ClassFanOutComplexity` max property from 20 to 21.

## Motivation
A recent build failure was detected in xwiki-platform-livedata-livetable where `LiveTableLiveDataConfigurationResolver.java` has a ClassFanOutComplexity of 21, exceeding the current max allowed value of 20.

## Details
- Changed the max value from 20 to 21 in the checkstyle configuration
- This change is needed as a notification that the fixed rule will raise violations after you upgrade to the latest Checkstyle version

## Implementation Notes
This PR should be merged after we release the new Checkstyle version and xwiki upgrades to it. Without this change, existing code that was previously compliant may fail validation checks.

## Testing Done
Verified that the updated configuration resolves the build error while maintaining code quality standards.

### DIFF
--- a/xwiki-commons-tools/xwiki-commons-tool-verification-resources/src/main/resources/checkstyle.xml
+++ b/xwiki-commons-tools/xwiki-commons-tool-verification-resources/src/main/resources/checkstyle.xml
@@ -136,6 +136,7 @@
         ComponentInstantiationStrategy, Unstable, Disposable, Initializable, InitializationException"/>
       <property name="excludedPackages"
         value="org.apache.maven.plugins.annotations, javax.annotation.concurrent, org.xwiki.component.annotation"/>
+      <property name="max" value="21"/>
     </module>
 
     <module name="ClassTypeParameterName"/>


michitux:
I think we would rather solve that with a `@SupressWarnings` annotation on the affected class than to increase the limit globally, if it shouldn't be possible to change the class to respect the current limit. Do you have any insights why CheckStyle will report a higher fan-out complexity? Which classes are considered that aren't considered right now?

Anmol202005:
https://github.com/xwiki/xwiki-platform/blob/92df84a3a91bb6dea21adcd35d3ce440757d0a2b/xwiki-platform-core/xwiki-platform-livedata/xwiki-platform-livedata-livetable/src/main/java/org/xwiki/livedata/internal/livetable/LiveTableLiveDataConfigurationResolver.java#L216

`.map(Constraint::new).collect(Collectors.toList())`

Previously, the check ignored Constraint::new in method references when calculating class fan-out complexity. This was a false negative. Now fixed, the check correctly counts Constraint::new as a dependency, which is why complexity is increased to 21.

tmortagne:
> I think we would rather solve that with a @SupressWarnings annotation on the affected class than to increase the limit globally

Yes, this PR is definitely not the right way to deal with this situation. We are not going to change the general rule for this specific case.

romani:
@tmortagne , do you prefer suppression by annotation ? We can change PR to suppress all violations due to this defect fix, to let your team to decided later on better approach.

romani:
@Anmol202005, please change PR to use suppressions.

PR #1261, opened 2025-03-05
XCOMMONS-3276: NPE Warnings after upgrading XWiki to 16.10.3 related to DefaultJobStatusStore - Failed to load job status

michitux:
# Jira URL

* https://jira.xwiki.org/browse/XCOMMONS-3276
* https://jira.xwiki.org/browse/XCOMMONS-3275

# Changes

## Description

* Use the existing encoding that handles null elements in the job id as before XCOMMONS-2387.
* Increment the version to migrate all job statuses to the new version, fixing XCOMMONS-3275.
* Fix creating the last part of the directory tree.
* Add and improve tests.

## Clarifications

* I've handled more here than in the original Jira issue as it didn't seem good to me to fix this one regression without fixing the others. 
* I've reverted the job status storage to use just URL encoding as it is way easier to inspect as a human. Further, it means that for existing installations that upgrade directly to this version no job status needs to be moved (and for all others, only job statuses created since the upgrade need to be moved).
* I haven't tested the upgrade. I think we sometimes override job statuses when we re-execute a job. I'm wondering if the `repair` code would need some bulletproofing for this case. From my understanding, it would log an error in this case. We might not want to do this and instead remove the old job status, but that seems specific to the current situation where a log might exist both without and with base64 encoding, so I'm not sure if this is a good idea generically.
* I'm wondering how we could protect against future changes to the folder structure without incrementing the `VERSION`. Maybe we could save a specific job status in all upgrade tests and check if it can still be retrieved after the upgrade?

# Screenshots & Video

No UI changes.

# Executed Tests

In `xwiki-commons`:
```
LANG=C.UTF-8 mvn clean install -Pdocker,legacy,integration-tests,snapshotModules,quality,distribution,flavor-integration-tests,standalone
```

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * stable-16.10.x


xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/DefaultJobStatusStore.java
```diff
@@ -211,13 +209,11 @@ private String encode(String name)
...
-            try {
-                encoded = URLEncoder.encode(name, DEFAULT_ENCODING);
-            } catch (UnsupportedEncodingException e) {
-                // Should never happen
-
-                encoded = name;
-            }
+            encoded = URLEncoder.encode(name, StandardCharsets.UTF_8)
+                // Replace characters that might be problematic in file systems.
+                .replace("+", "%20")
```
tmortagne, line 214: Isn't URLEncoder already supposed to encode the `+` characters as "%2B" ?

michitux: Yes, but it encodes spaces as `+`.

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/DefaultJobStatusStore.java
```diff
@@ -211,13 +209,11 @@ private String encode(String name)
...
+                .replace(".", "%2E")
+                .replace("*", "%2A");
```
tmortagne, line 216: `FileSystemStoreUtils#encode` is a bit more complete and focuses on this kind of stuff (while still being based on URL encoding), but it encodes more (to avoid collisions in case-insensitive filesystems it encodes uppercase characters, for example).

tmortagne:
Ideally we should probably refactor how `repair()` is implemented, as it's quite expensive right now (it loads the entire job status just to get the id basically).

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/DefaultJobStatusStore.java
tmortagne: Right, not a very accurate encoding...

michitux: In the end, it's probably fine to leave the `+` characters as-is, it's just not a very good encoding (and doesn't work in URLs in some cases, see https://jira.xwiki.org/browse/XWIKI-22956).

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/DefaultJobStatusStore.java
michitux: Indeed, it probably makes sense to use this here, too, so we have one central place to fix if there should be any problem.

michitux:
> Ideally we should probably refactor a bit how `repair()` is implemented, as it's quite expensive right now (it loads the entire job status just to get the id basically).

Indeed, and ideally we should also add some handling to resolve conflicts, meaning that the target directory already contains a job status, by keeping the more recent job status.

tmortagne:
> Maybe we could save a specific job status in all upgrade tests and check if it can still be retrieved after the upgrade?

We don't necessarily need to have it in all upgrade tests, but at least xwiki-platform-distribution-flavor-test-upgrade-1210-custom (which is already designed to be based on custom data) would be nice for example.

michitux:
> We don't necessarily need to have it in all upgrade tests, but at least xwiki-platform-distribution-flavor-test-upgrade-1210-custom (which is already designed to be based on custom data) would be nice for example.

That might not be enough, by upgrading directly to the current version, we won't detect if the format is changed again. I think we need to have a new one at least after every format change to be able to verify that the format was incremented again afterwards.

michitux:
> it loads the entire job status just to get the id basically

Do you have any suggestions how to make it less expensive?

tmortagne:
> Do you have any suggestions how to make it less expensive?

All we can do is add a special helper which parse the status XML but ignore everything which is not the id (mainly the progress) and stops as soon as it finds it. It will still read more than needed, but just reading characters and ignoring them should still be faster than interpreting the whole status.

tmortagne:
> That might not be enough, by upgrading directly to the current version, we won't detect if the format is changed again. I think we need to have a new one at least after every format change to be able to verify that the format was incremented again afterwards.

It is enough now. I agree with you that when the format changes again we should add another one, but there is no much we can do about it now.

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/DefaultJobStatusStore.java
michitux: I just noticed that `FileSystemStoreUtils` is in `xwiki-platform`… Any suggestions where to move it in `xwiki-commons`?

tmortagne: Hmm, I guess we could either move it to xwiki-commons-text, or start a xwiki-commons-store module (we will probably have more stuff in there, thinking for example about the plan to provide an API to ask as a proxy to store files either in the filesystem or an object storage).

michitux:
@tmortagne I'm not really happy with it, but I've added a huge chunk of code for extracting the ID of a job which might be faster than simply unserializing the job status.

michitux:
> Indeed, and ideally we should also add some handling to resolve conflicts, meaning that the target directory already contains a job status, by keeping the more recent job status.

Any suggestions how we should define "more recent" - should we parse the job status or should we just use the last modified (or created?) time of the file system?

tmortagne:
>  the last modified (or created?) time of the file system?

Sounds like this should be good enough.

What do you plan to do with the "looser", keep it at the old location or delete it?

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/DefaultJobStatusStore.java
michitux: I've moved it to `xwiki-commons-text` for now as it feels strange to create a new module for an internal class. We can move it later when implementing that API.

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/DefaultJobStatusStore.java
michitux: The code is using `FileSystemStoreUtils#encode` which keeps `+` as encoding of spaces.

michitux:
After writing all this code, I'm now wondering if this isn't too big and if we really want to have this huge change.

An idea how to make this much simpler:

* Keep all versions of encode (the original one, the one from before this PR, and the proposed one).
* When loading a job status, check all locations, starting with the current one.
* If several locations exist, choose the one with the most recent job status.
* Don't perform any repair (but keep the improvements to move the log).
* Remove the huge chunk of code for reading just the job ID again.

Where I'm hesitating is the case where a job status is loaded and saved again, I'm wondering if we wouldn't lose the log in this case. So maybe that's again not the solution? Do you have any other suggestions how to simplify this?

xwiki-commons-core/xwiki-commons-text/src/main/java/org/xwiki/text/internal/FileSystemStoreUtils.java
```diff
@@ -0,0 +1,137 @@
...
+ * 
+ * @version $Id$
+ * @since 10.1RC1
+ */
+public final class FileSystemStoreUtils
+{
+    /**
+     * The standard role hint used by most filesystem store implementations.
+     */
+    public static final String HINT = "file";
```
surli, line 38: Feels like something you moved? I don't see where the HINT is used in your changes? And the since is wrong?

michitux: Indeed, I moved the file without paying too much attention to that, good catch. Now this hint is used quite a lot on the `xwiki-platform` side, so I'm not sure what to do. Move it into a separate class? This is for example used in https://github.com/xwiki/xwiki-platform/blob/7dee366a67352adc78da6aef6bc67c867818414a/xwiki-platform-core/xwiki-platform-store/xwiki-platform-store-filesystem-oldcore/src/main/java/org/xwiki/store/legacy/store/internal/FilesystemAttachmentStore.java#L231

xwiki-commons-core/xwiki-commons-text/src/main/java/org/xwiki/text/internal/FileSystemStoreUtils.java
```diff
@@ -0,0 +1,137 @@
...
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
+
+import org.apache.commons.lang3.CharUtils;
+
+/**
+ * Various file system module utils.
+ * 
+ * @version $Id$
+ * @since 10.1RC1
```
tmortagne, line 31: Wrong since.

tmortagne:
>  An idea how to make this much simpler:

This kind of mean getting rid of the concept of repair and storage version (since you end up with a mixed storage).

It feels like a good idea, but it's not really clear to me what is the problem you have with the log. If you move everything from the old folder to the new one before loading it it should be fine.

xwiki-commons-core/xwiki-commons-text/src/main/java/org/xwiki/text/internal/FileSystemStoreUtils.java
surli: It's internal only, so if you change the imports in xwiki-platform to use that file from that module I'd say it's ok. You just need to prepare the PR in platform for that and link it here to understand the changes I'd say.

michitux:
> It feels like a good idea, but it's not really clear to me what is the problem you have with the log, so I'm probably missing something. If you move everything from the old folder to the new one before loading it, it should be fine.

I didn't suggest moving the job status after loading it, but it could indeed be a possibility. I was thinking of a flow like load the job status, modify a property, save it again. And I'm not sure where in that flow we would move everything. Now even if we move the job status directly after/before loading, it could indeed be an interesting optimization as we would move it at a point where we parse the whole job status, anyway, so we could avoid the need for the ID parsing optimization.

To me, it just feels dangerous to move files around, in particular if we delete files during that move, that just feels like the kind of feature where we can easily lose the whole data, because, e.g., we wrongly read the ID and moved everything to the same directory.

xwiki-commons-core/xwiki-commons-text/src/main/java/org/xwiki/text/internal/FileSystemStoreUtils.java
tmortagne: But it does not make much sense right now to get this metadata from `xwiki-commons-text`. Would be better to put HINT back to platform, probably somewhere in `xwiki-platform-store-filesystem`.

michitux:
> Then just copy it, but it feels a bit like a pity to have those leftovers (now there should not be that many with the new format). IMO, it's better to put the file in the right location so that next time it's found right away, and it's also easier to find for a human.

I don't think copying is a solution. The main thing I worry about is deleting files, not so much moving them. The main question to me is what should happen and how we should detect if a job exists both at the new location and an old location, and the old location is the more recent one (as we kind of moved back to almost the old location). We could also say we ignore that by taking the first location we find and only moving files when the new location doesn't have any, and it's an expected data loss that if you used 16.10.x/17.x before the fix, you'll just get the older job status and you'll have some leftover files.

tmortagne:
> We could also say we ignore that by taking the first location we find and only moving files when the new location doesn't have any

That's definitely what I have in mind, I really don't think it worth doing anything more complex than that.

xwiki-commons-core/xwiki-commons-text/src/main/java/org/xwiki/text/internal/FileSystemStoreUtils.java
michitux: I've analyzed a bit how the `encode` method of this class is used in `xwiki-platform` and I came to the conclusion that moving and re-using this class seems wrong. This `encode` method is solely used for moving attachments that were previously stored with this kind of encoding in a migration. This means that we must not change it in any way, or the migration would fail. Even if we should notice that the encoding causes problems in certain scenarios, we wouldn't change it. While we need stable encoding here, too, I believe it's a bad idea to re-use this class as here, we would change the implementation if we noticed any problems.

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/Version3JobStatusFolderResolver.java
```diff
@@ -0,0 +1,107 @@
...
+ * Version 3 of {@link JobStatusFolderResolver} that uses URL encoding but with length limit and additional encoding
+ * for reserved characters. Introduced in XWiki 16.10.6 and 17.2.0.
+ *
+ * @version $Id$
+ * @since 17.2.0RC1
+ * @since 16.10.6
+ */
+@Component
+@Singleton
+@Named("version3")
```
tmortagne, line 43: I'm not sure we really need to put a prefix here. The number as a String should be fine.

michitux: We need some prefix in the class name and to me, it felt more consistent to have the same prefix in the hint.

tmortagne (approved):

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/DefaultJobStatusStore.java
```diff
@@ -378,7 +406,8 @@ private void saveJobStatus(JobStatus status)
             this.writeLock.lock();
 
             try {
-                File statusFile = getJobFolder(status.getRequest().getId());
+                // FIXME: not clear if the job status should be moved here.
```
michitux, line 409: @tmortagne Have you seen this FIXME?

michitux:
I've just tested these changes on an instance that had job statuses of extension installations from before this PR (17.2.0-SNAPSHOT from a couple of days ago) and it successfully moved the job statuses to the new directory structure. The old directories were completely deleted after accessing the progress of the extension installations.

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/DefaultJobStatusStore.java
tmortagne: I indeed missed it. It's not clear to me why it's a `FIXME`, actually.

You are wondering if it's right to move the status just to immediately overwrite it ?

In quite a lot of use cases, that id will go through get getJobStatus (which does the move, from what I understand) before. So I don't see why those use cases where there is no getJobStatus before the save should have a different end result, if anything at least for consistency.

michitux: Yes, okay, I've removed the FIXME. I agree, I just wasn't 100% sure.

PR #1325, opened 2025-04-29
XCOMMONS-3327: Provide an internal helper to safely access ClassLoader resources

tmortagne:
# Changes

## Description

The goal of this pull request is to provide an internal helpers that prevent traversal attacks when accessing resources from the classloader.

# Executed Tests

The pull request comes with unit tests.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches: 16.4.x, 16.10.x

### DIFF
--- a/xwiki-commons-core/xwiki-commons-classloader/xwiki-commons-classloader-api/src/main/java/org/xwiki/classloader/internal/ClassLoaderUtils.java
+++ b/xwiki-commons-core/xwiki-commons-classloader/xwiki-commons-classloader-api/src/main/java/org/xwiki/classloader/internal/ClassLoaderUtils.java
@@ -1,0 +1,142 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.xwiki.classloader.internal;
+
+import java.io.InputStream;
+import java.net.URL;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * Various {@link ClassLoader} tools to make easier to safely access resources.
+ * 
+ * @version $Id$
+ * @since 17.4.0RC1
+ * @since 16.10.7
+ */
+public final class ClassLoaderUtils
+{
+    private ClassLoaderUtils()
+    {
+
+    }
+
+    private static String resolveResourceName(String prefixPath, String resourcePath)
+    {
+        String fullPath;
+        if (StringUtils.isEmpty(prefixPath)) {
+            fullPath = resourcePath;
+
+            // Prevent access to resources from other directories
+            Path normalizedResource = Paths.get(fullPath).normalize();
+            if (normalizedResource.startsWith("../")) {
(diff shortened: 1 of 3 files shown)


xwiki-commons-core/xwiki-commons-classloader/xwiki-commons-classloader-api/src/test/java/org/xwiki/classloader/internal/ClassLoaderUtilsTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+    private static final String RESOURCE_NAME_BACK = "folder/../" + RESOURCE_NAME;
+
+    private static final String PREFIXED_RESOURCE_NAME_BACK = PREFIX + RESOURCE_NAME_BACK;
+
+    private ClassLoader classLoader;
+
+    private URL resouceURL;
+
+    private URL prefixedResouceURL;
```
manuelleduc, line 55: I might be missing something, but I can't find where this field is initialized.

xwiki-commons-core/xwiki-commons-classloader/xwiki-commons-classloader-api/src/main/java/org/xwiki/classloader/internal/ClassLoaderUtils.java
```diff
@@ -0,0 +1,139 @@
...
+            // Prevent access to resources from other directories
+            Path normalizedResource = Paths.get(fullPath).normalize();
+            if (!normalizedResource.startsWith(prefixPath)) {
+                throw new IllegalArgumentException(String.format(
+                    "The provided resource name [%s] is trying to navigate out of the mandatory prefix [%s]",
+                    resourcePath, prefixPath));
+            }
+        }
+
+        return fullPath;
```
michitux, line 69: To me, it would seem safer to return the normalized path, not the original path. This would avoid the exploitation of any parsing differences between `Path` and the class loader. I don't know if there are any, but I know that it's a thing for URL parsing, see, e.g., https://claroty.com/team82/research/exploiting-url-parsing-confusion. Citing https://docs.oracle.com/javase/8/docs/technotes/guides/lang/resources.html:

> The getResource and getResourceAsStream methods find a resource with a given name. They return null if they do not find a resource with the specified name. The rules for searching for resources associated with a given class are implemented by the class's ClassLoader. The Class methods delegate to ClassLoader methods, after applying a naming convention: if the resource name starts with "/", it is used as is. Otherwise, the name of the package is prepended, after converting all periods (.) to slashes (/).

To me this sounds as if there might be significant parsing differences between `Path` and loading resources. If we want to avoid that, we should ensure that we normalize the path as much as possible (like, ensure that it always starts with `/`).

xwiki-commons-core/xwiki-commons-classloader/xwiki-commons-classloader-api/src/main/java/org/xwiki/classloader/internal/ClassLoaderUtils.java
```diff
@@ -0,0 +1,139 @@
...
+
+import java.io.InputStream;
+import java.net.URL;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * Various {@link ClassLoader}.
```
manuelleduc, line 30: unfinished sentence?

xwiki-commons-core/xwiki-commons-classloader/xwiki-commons-classloader-api/src/test/java/org/xwiki/classloader/internal/ClassLoaderUtilsTest.java
michitux:
```suggestion
    private URL prefixedResourceURL;
```

Indeed, I also don't see any initialization. Further, there is a typo (needs to be properly renamed).

xwiki-commons-core/xwiki-commons-classloader/xwiki-commons-classloader-api/src/main/java/org/xwiki/classloader/internal/ClassLoaderUtils.java
tmortagne: I hesitated and decided to keep the original path to avoid causing regressions. But I agree with you that it's safer.

tmortagne: > But I agree with you that it's safer.

Turns out, its behavior is a bit too native on Windows, and it replaces all `/` by `\` (windows does not play well with the Classloader).

PR #1329, opened 2025-05-09
XCOMMONS-3330: Job groups with several threads may not return a current job

michitux:
# Jira URL

https://jira.xwiki.org/browse/XCOMMONS-3330

# Changes

## Description

* Store several current jobs.
* Add a new possibility to return all running jobs of a job group.
* Deprecate the old method that returns just a single running job.
* Clean up empty entries in groupedJobs.
* Some code cleanup.

## Clarifications

* This needs more testing, in particular in automated tests.

# Screenshots & Video

No UI changes.

# Executed Tests

```
mvn clean install -Pintegration-tests,docker,legacy,quality -pl :xwiki-commons-job-api,:xwiki-commons-job-default
```

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * Not sure, we could decide to backport this.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/DefaultJobExecutor.java
+++ b/xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/DefaultJobExecutor.java
@@ -133,25 +133,22 @@ public class DefaultJobExecutor implements JobExecutor, Initializable, Disposabl
 
             List<String> jobId = job.getRequest().getId();
             if (jobId != null) {
-                // Synchronize on groupedJobs to avoid race conditions with jobs completing before they are inserted
-                // into the queue.
-                synchronized (DefaultJobExecutor.this.groupedJobs) {
-                    // Delete the job from the job group's queue. Remove the queue when it is empty.
-                    DefaultJobExecutor.this.groupedJobs.compute(jobId, (k, v) -> {
-                        // Remove the job, but only when it is the first job in the queue.
-                        if (v != null && v.peek() == job) {
-                            v.poll();
-                        }
+                // Delete the job from the job group's queue. Remove the queue when it is empty.
+                // Use compute for synchronization.
+                DefaultJobExecutor.this.groupedJobs.compute(jobId, (k, v) -> {
+                    // Remove the job, but only when it is the first job in the queue.
+                    if (v != null && v.peek() == job) {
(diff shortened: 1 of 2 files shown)


xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/DefaultJobExecutor.java
```diff
@@ -262,11 +274,23 @@ public void dispose() throws ComponentLifecycleException
...
+
+    @Override
+    public Collection<Job> getCurrentJobs(JobGroupPath path)
     {
         JobGroupExecutor executor = this.groupExecutors.get(path);
 
-        return executor != null ? executor.currentJob : null;
+        if (executor != null) {
+            // Return an unmodifiable copy of the set of currently running jobs.
+            return Collections.unmodifiableCollection(new ArrayList<>(executor.currentJobs));
```
tmortagne, line 290: It's not clear to me why you create an ArrayList. Isn't Collections.unmodifiableCollection enough to make the set readonly ?

tmortagne (approved):

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/DefaultJobExecutor.java
michitux: I want to avoid that the collection updates when jobs are started or stopped. I would expect that we expose this method in a script service and I don't want to let the author of a Velocity script deal with the fact that directly after `isEmpty()` returns `false`, there might not be any elements anymore in the collection.

We could also let this return a `List`, this might make things more explicit what to expect in terms of the different operations?

tmortagne: > I want to avoid that the collection updates when jobs are started or stopped.

Very good point indeed.

> We could also let this return a List, this might make things more explicit what to expect in terms of the different operations?

Why not.

michitux: Done. I also found and fixed some issues in the implementations for constructing the list both in the default and in the actual implementation.

PR #1381, opened 2025-06-30
[Misc] add import order for jakarta

manuelleduc:
Fix import order for jakarta.


tmortagne:
I thought I had done it in https://github.com/xwiki/xwiki-commons/commit/00e64de09367e37b2c49db1d893cb556d55b69fc (and asked IntelliJ users to check it back then), but apparently the imports are declared several times.

PR #1398, opened 2025-07-16
XCOMMONS-3382: Upgrade to commons-lang3 3.18.0

michitux:
# Jira URL

https://jira.xwiki.org/browse/XCOMMONS-3382

# Changes

## Description

* Un-deprecate the newly deprecated methods as we offer no alternative in Velocity, and the new Strings API seems very confusing so far as it's inconsistent which methods are in `Strings` and which are still in `StringUtils`.

## Clarifications

* I've copied the documentation to have nice documentation in the scripting API.
* This is a security update. The security vulnerability shouldn't really affect XWiki, but it's still better to upgrade in case an extension is using one of the vulnerable methods. For this reason, it should also be backported on the LTS branches.

# Screenshots & Video

No UI changes.

# Executed Tests

```
mvn clean install -Pquality,legacy -pl :xwiki-commons,:xwiki-commons-velocity,:xwiki-commons-legacy-velocity
```

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * stable-17.4.x
  * stable-16.10.x

### DIFF
--- a/xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/StringTool.java
+++ b/xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/StringTool.java
@@ -19,19 +19,25 @@
  */
 package org.xwiki.velocity.tools;
 
+import java.util.Objects;
+
+import org.apache.commons.lang3.Strings;
 import org.xwiki.text.StringUtils;
 
 /**
  * Velocity Tool providing various helpers to manipulate strings.
- * 
+ *
+ * <p>This class is basically {@link org.apache.commons.lang3.StringUtils} but with fewer deprecated methods, to
+ * control the deprecation in XWiki and to keep methods available in Velocity whose recommended alternative isn't
+ * available in Velocity.</p>
+ *
  * @version $Id$
  * @since 15.7RC1
  */
 public class StringTool extends StringUtils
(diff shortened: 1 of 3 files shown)


xwiki-commons-core/xwiki-commons-velocity/src/main/java/org/xwiki/velocity/tools/StringTool.java
```diff
@@ -41,6 +47,1229 @@ public class StringTool extends StringUtils
...
+     * StringTool.appendIfMissing("abcXYZ", "xyz", "mno") = "abcXYZxyz"
+     * StringTool.appendIfMissing("abcMNO", "xyz", "mno") = "abcMNOxyz"
+     * </pre>
+     *
+     * @param str      The string.
+     * @param suffix   The suffix to append to the end of the string.
+     * @param suffixes Additional suffixes that are valid terminators.
+     * @return A new String if suffix was appended, the same string otherwise.
+     */
+    public static String appendIfMissing(final String str, final CharSequence suffix, final CharSequence... suffixes)
```
surli, line 86: Sounds like you forgot the unstable and since everywhere in this?

surli: Ah. Good point indeed.

PR #1416, opened 2025-08-05
XCOMMONS-3250: Make as easy as possible for various features to use an object store

michitux:
# Jira URL

https://jira.xwiki.org/browse/XCOMMONS-3250

# Changes

## Description

* Introduce a blob storage API with implementations for file system and Amazon S3.

## Clarifications

* This is an initial draft, the S3 version is probably missing some essential configuration options for third-party providers.
* The code is currently untested and also missing unit tests.
* Before this is merged/released, I think we should migrate at least one use case in XWiki to this new blob storage to see if the API actually works.
* This PR exists to provide visibility into the ongoing work and to get feedback on the API.

# Screenshots & Video

No UI changes.

# Executed Tests

None so far.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * None


xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-api/src/main/java/org/xwiki/store/blob/Blob.java
```diff
@@ -0,0 +1,113 @@
...
+import java.io.OutputStream;
+
+import org.xwiki.stability.Unstable;
+import org.xwiki.store.StreamProvider;
+
+/**
+ * A Blob is a piece of data stored in a BlobStore.
+ *
+ * @version $Id$
+ * @since 17.7.0RC1
```
tmortagne, line 32: I guess it's not going to be the case, but anyway, hard to be sure what to put yet.

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-api/src/main/java/org/xwiki/store/blob/Blob.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * Write the content of the given InputStream to this blob.
+     *
+     * @param inputStream the InputStream to read data from
+     * @param conditions the conditions that must be satisfied before writing to this blob
+     * @throws BlobStoreException if the InputStream cannot be read or the blob cannot be written, for example because
+     * its name is invalid.
+     */
+    void writeFromStream(InputStream inputStream, WriteCondition... conditions) throws BlobStoreException;
```
tmortagne, line 78: You think there are cases when getOutputStream would not be enough or that passing the inputstream to write would allow some optimizations ? As otherwise, I don't see much point of exposing this API, even as a helper IOUtils#copy should be good enough.

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-api/src/main/java/org/xwiki/store/blob/BlobPath.java
```diff
@@ -0,0 +1,143 @@
...
+        if (!(o instanceof BlobPath other)) {
+            return false;
+        }
+        return this.raw.equals(other.raw);
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(this.raw);
```
tmortagne, line 141: Sounds simpler and faster to just use `this.raw.hasCode()`

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-api/src/main/java/org/xwiki/store/blob/BlobPath.java
```diff
@@ -0,0 +1,143 @@
...
+    public static BlobPath of(List<String> segments)
+    {
+        if (segments.isEmpty()) {
+            throw new IllegalArgumentException("At least one segment required");
+        }
+        return new BlobPath(segments);
+    }
+
+    /**
+     * Create a BlobPath by splitting a slash-delimited string.
```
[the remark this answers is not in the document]
michitux: It's not probably not clear enough, but the idea wasn't that this API would be used with user-supplied paths. The idea was really that this is an internal API that is used with already "safe" paths and just replaces using raw filesystem access.

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-filesystem/src/main/java/org/xwiki/store/blob/internal/FileSystemBlobStoreManager.java
```diff
@@ -0,0 +1,54 @@
...
+@Named("filesystem")
+public class FileSystemBlobStoreManager implements BlobStoreManager
+{
+    @Inject
+    private Environment environment;
+
+    @Override
+    public BlobStore getBlobStore(String name) throws BlobStoreException
+    {
+        Path basePath = this.environment.getPermanentDirectory().toPath().resolve(name);
```
tmortagne, line 51: Hmm, I feel safer to put those stores in a sub folder, like `store/` (which already exist, but does not currently contain any `filesystem` store, the current file store is in `file`).

We could still have stuff in the permanent directory which we don't want to share with other cluster member but which are still not something that would go in the temporary directory.

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-s3/src/main/java/org/xwiki/store/blob/internal/S3Blob.java
```diff
@@ -0,0 +1,160 @@
...
+import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
+import software.amazon.awssdk.services.s3.model.S3Exception;
+
+/**
+ * A {@link Blob} implementation that represents a blob stored in S3.
+ *
+ * @version $Id$
+ * @since 17.7.0RC1
+ */
+public class S3Blob implements Blob
```
tmortagne, line 47: A few things might worth an `AbstractBlob` (the `path` and `store` fields, with a generic, `#getStore()`, `#getPath`, some default implementation of `#writeFromStream` if we keep it), to reduce some noise at actual implementation level.

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-api/src/main/java/org/xwiki/store/blob/BlobStore.java
```diff
@@ -0,0 +1,92 @@
...
+import org.xwiki.stability.Unstable;
+
+/**
+ * A storage that allows storing blob data.
+ *
+ * @version $Id$
+ * @since 17.7.0RC1
+ */
+@Unstable
+public interface BlobStore
```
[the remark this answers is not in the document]
michitux: Indeed, though for S3 this will be just copy + delete from what I can see. For the filesystem backend it probably makes sense to have a better implementation.

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-api/src/main/java/org/xwiki/store/blob/BlobStoreManager.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Blob store manager for getting or creating blob stores.
+ *
+ * @version $Id$
+ * @since 17.7.0RC1
+ */
+@Role
+@Unstable
+public interface BlobStoreManager
```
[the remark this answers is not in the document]
michitux: Good point, though at the moment it's not really needed. The only important thing regarding disposing/closing right now is to close all input streams that are returned from an `S3Blob` in a timely manner as each of them will hold an open HTTP connection and thus block one connection of the connection pool.

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-api/src/main/java/org/xwiki/store/blob/BlobPath.java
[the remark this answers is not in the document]
tmortagne: Well, right now it's not internal at all (it's a public API). I also think that it could be interesting to have a generic syntax to allow us to serialize the `BlobPath` in case we need to refer to it somewhere else (the database, etc).

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-filesystem/src/main/java/org/xwiki/store/blob/internal/FileSystemBlobStoreManager.java
tmortagne: Very good point, indeed. Keeping the exact same files and just change the API to access them would definitely make upgrades a lot easier.

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-api/src/main/java/org/xwiki/store/blob/Blob.java
michitux: For now, I've added a comment to reconsider the method.

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-api/src/main/java/org/xwiki/store/blob/Blob.java
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * Copy this blob to another path within the same store.
+     *
+     * @param targetPath the target path for the copy
+     * @return the new blob at the target path
+     * @throws BlobStoreException if the copy operation fails, for example because this blob doesn't exist
+     * @throws BlobNotFoundException if this blob does not exist
+     * @throws BlobAlreadyExistsException if a blob already exists at the target path
+     */
+    default Blob copyTo(BlobPath targetPath) throws BlobStoreException
```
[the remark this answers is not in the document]
michitux: I've removed them.

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-api/src/main/java/org/xwiki/store/blob/BlobPath.java
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * @version $Id$
+ * @since 17.7.0RC1
+ */
+@Unstable
+public final class BlobPath
+{
+    private final List<String> segments;
+
+    private final String raw;
```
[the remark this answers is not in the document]
michitux: I've renamed the field.

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-api/src/main/java/org/xwiki/store/StreamProvider.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.xwiki.store;
+
+import java.io.InputStream;
+
+/**
+ * A generic thing which provides access to an InputStream on demand.
+ *
+ * @version $Id$
+ * @since 3.0M2
```
manuelleduc: Given the odd version I guess it's a moved file but pointing it out since github is not marking the file as moved.

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-api/src/main/java/org/xwiki/store/StoreException.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.io.Serial;
+
+import org.xwiki.stability.Unstable;
+
+/**
+ * Base exception for store related APIs.
+ *
+ * @version $Id$
+ * @since 17.1.0RC1
```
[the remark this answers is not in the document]
michitux: See https://github.com/xwiki/xwiki-platform/blob/566248b59d81ca47377573ab6a58ae2a538b46e6/xwiki-platform-core/xwiki-platform-store/xwiki-platform-store-api/src/main/java/org/xwiki/store/StoreException.java

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-api/src/main/java/org/xwiki/store/blob/Blob.java
michitux: I've just noticed that I still haven't resolved this. So the current status is:

* There is still no implementation that benefits from `writeFromStream`. I've decided against the async client as it would introduce a lot of overhead and boilerplate code for currently almost no benefit (except maybe for uploading large files).
* In the calling code, `writeFromStream` seems quite convenient and is used in a few places.

So I'm still not sure what to do here.

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-api/src/main/java/org/xwiki/store/blob/BlobStore.java
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * List all descendant blobs under the given path prefix. The returned stream includes blobs at every depth below
+     * the provided path. The caller must close the returned stream after use.
+     *
+     * @param path the path prefix to search under
+     * @return a stream of the blobs under the given path
+     * @throws BlobStoreException if the listing operation fails
+     */
+    Stream<Blob> listDescendants(BlobPath path) throws BlobStoreException;
```
tmortagne: I would have also expected an API to get only direct descendants, sounds like a common use case.

xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-api/src/main/java/org/xwiki/store/blob/BlobStoreFactory.java
```diff
@@ -0,0 +1,64 @@
...
+     * @param properties the populated properties bean to use
+     * @return a BlobStore instance
+     * @throws BlobStoreException if the properties are invalid or creation fails
+     */
+    BlobStore create(String name, BlobStoreProperties properties) throws BlobStoreException;
+
+    /**
+     * @return the properties class used by this factory
+     */
+    Class<? extends BlobStoreProperties> getPropertiesClass();
```
tmortagne: If `BlobStoreFactory` is associated to a child of `BlobStoreProperties`, wouldn't it make more sense to make it more explicit and declare that child as a generic of the class and use that generic in `#create` ?

michitux: I've implemented this in 3a557767b4b234f35cd74bafe602988d9dc95406.

PR #1651, opened 2026-02-25
[Misc] Capture only logs from specified logger

michitux:
# Jira URL

None (test-only change).

# Changes

## Description

* Add a feature to LogCaptureExtension to only capture logs from a specified logger.
* Add a unit test to LogCaptureExtensionTest.
* Use the new feature in S3 blob store tests to avoid capturing debug logs from background threads.

## Clarifications

* I'm opening this PR mainly to get a quick feedback if the logger setup looks okay or if this could introduce any problems.

# Screenshots & Video

No UI changes.

# Executed Tests

Built the changed modules with the quality profile.

# Expected merging strategy

* Prefers squash: Yes 
* Backport on branches:
  * stable-17.10.x.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-s3/src/test/java/org/xwiki/store/blob/internal/S3MultipartUploadHelperTest.java
+++ b/xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-s3/src/test/java/org/xwiki/store/blob/internal/S3MultipartUploadHelperTest.java
@@ -76,8 +76,11 @@ class S3MultipartUploadHelperTest
     @Mock
     private BlobPath blobPath;
 
+    // Capture debug logs only from our own package to avoid capturing debug logs of background threads that might
+    // still be running from previously executed tests and that would interfere with assertions on the logs in this
+    // test class.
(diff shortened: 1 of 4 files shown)


xwiki-commons-core/xwiki-commons-store/xwiki-commons-store-blob/xwiki-commons-store-blob-s3/src/test/java/org/xwiki/store/blob/internal/S3MultipartUploadHelperTest.java
```diff
@@ -76,8 +76,11 @@ class S3MultipartUploadHelperTest
     @Mock
     private BlobPath blobPath;
 
+    // Capture debug logs only from our own package to avoid capturing debug logs of background threads that might
+    // still be running from previously executed tests and that would interfere with assertions on the logs in this
+    // test class.
     @RegisterExtension
-    private LogCaptureExtension logCapture = new LogCaptureExtension(LogLevel.DEBUG);
+    private LogCaptureExtension logCapture = new LogCaptureExtension(LogLevel.DEBUG, "org.xwiki.store.blob");
```
manuelleduc: You confirm that with this new parameter, if the tested code produces unexpected logs with another logger name, they will be missed?

PR #1703, opened 2026-04-11
XWIKI-10330: Allow reinstalling an extension from an extension repository in the extension manager

vmassol:
# Jira URL

https://jira.xwiki.org/browse/XWIKI-10330

# Changes

## Description

This is the change in commons. There's also a PR for platform.

Coded by Claude Code. To be reviewed/validated.

# Executed Tests

Manual testing in EM to reinstall an extension.

# Expected merging strategy

* Prefers squash: Yes

### DIFF
--- a/xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/job/internal/AbstractInstallPlanJob.java
+++ b/xwiki-commons-core/xwiki-commons-extension/xwiki-commons-extension-api/src/main/java/org/xwiki/extension/job/internal/AbstractInstallPlanJob.java
@@ -1301,7 +1301,17 @@ public abstract class AbstractInstallPlanJob<R extends InstallRequest> extends A
         throws IncompatibleVersionConstraintException, ResolveException, InstallException
     {
         if (getRequest().isInstalledIgnored()) {
-            return Set.of();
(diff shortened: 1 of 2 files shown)


vmassol:
Closing this together with its platform counterpart xwiki/xwiki-platform#5383 — see [that discussion](https://github.com/xwiki/xwiki-platform/pull/5383#issuecomment-5410278130) for the reasoning.

Short version: @tmortagne is right that the SNAPSHOT dev loop I was actually after is an *upgrade*, not a *reinstall*, and is therefore XCOMMONS-2748's job rather than XWIKI-10330's. Framed as an upgrade, the XAR merge problem this PR was working around disappears on its own, because `previous` and `next` are then genuinely different versions.

The one part of this PR worth keeping is the `LocalExtensionStorage#removeExtension()` fix, which is an independent bug on master. It has been extracted into XCOMMONS-3750 / #1925, with the two extra findings that came out of looking at it properly:

* since the mis-targeted `Files.delete()` throws `NoSuchFileException`, `deleteExtensionFolderIfEmpty()` is never reached either — so XCOMMONS-3197 never actually worked, and every extension removal logs an error;
* simply fixing the delete target is not safe on its own: descriptors are also supported directly at the repository root, so the "version folder" can *be* the repository root, which the fixed code would then delete (and then its parent) once empty.

The `AbstractInstallPlanJob` change from this PR is dropped — it only existed to support the reinstall flow.

PR #1825, opened 2026-07-07
Remove JavadocStyleCheck from checkstyle configuration

gianmarcoschifone:
Removes `JavadocStyle` from the shared XWiki Checkstyle configuration because the check is being removed from Checkstyle in checkstyle/checkstyle#20582.

I tested replacing it with [`SummaryJavadoc`](https://checkstyle.org/checks/javadoc/summaryjavadoc.html), but that changes the current behavior and introduces many new violations in `xwiki-rendering`. For now this PR only removes the obsolete check.

`SummaryJavadoc` can still be considered later as an alternative for enforcing Javadoc summary checks, after fixing the existing violations.

### DIFF
--- a/xwiki-commons-tools/xwiki-commons-tool-verification-resources/src/main/resources/checkstyle.xml
+++ b/xwiki-commons-tools/xwiki-commons-tool-verification-resources/src/main/resources/checkstyle.xml
@@ -242,12 +242,6 @@
       <property name="accessModifiers" value="public"/>
     </module>
 
-    <module name="JavadocStyle">
-      <!-- Disabled since it's buggy, see
-           https://forum.xwiki.org/t/disable-the-html-check-from-the-javadocstyle-rule/9513 -->
-      <property name="checkHtml" value="false"/>
-    </module>
-
     <module name="JavadocType">
       <property name="scope" value="public"/>
       <property name="versionFormat" value="\$Id.*\$"/>


vmassol:
@gianmarcoschifone thx for the PR. Have you built at least xwiki-commons and xwiki-rendering with the change, to prove that it doesn't fail the build? Thx

gianmarcoschifone:
@vmassol Hi, thank you for advice.
I tested replacing `JavadocStyle` with `SummaryJavadoc`, but it is not safe for this configuration right now.

When `SummaryJavadoc` is enabled through the shared XWiki Checkstyle configuration, `xwiki-rendering` fails with 428 new Checkstyle violations, all reported as `Summary javadoc is missing. [SummaryJavadoc]`.

So I updated this PR to only remove `JavadocStyle` for now. Projects that want to keep enforcing Javadoc summary checks can consider enabling `SummaryJavadoc` separately, but that would require fixing the existing violations first.

vmassol:
@gianmarcoschifone I think that removing the `JavadocStyle` check is going to remove lots of checks that we had before. AFAIU we were only disabling `checkHtml` but the other settings that were true by default were verified (for example `[checkFirstSentence](https://checkstyle.sourceforge.io/checks/javadoc/javadocstyle.html#checkFirstSentence)` is true by default).

Thx

gianmarcoschifone:
@vmassol 
You're right that removing `JavadocStyle` removes more than just the HTML validation.

However, `JavadocStyle` has problematic and conceptually inconsistent behavior for summary validation. In this configuration it was checking punctuation only when it found free-form descriptive text, but it was not actually enforcing that every Javadoc has a summary. For example, Javadocs containing only block tags such as `@return`, `@param`, `@see`, `@version` or `@since` were not reported, because `checkEmptyJavadoc` was false by default.

I tested `SummaryJavadoc`, which is the proper check for summary validation, but it reports many existing violations in XWiki: 5 in `xwiki-commons` and 428 in `xwiki-rendering`.

So for now this PR only removes `JavadocStyle`, because the check is being removed from Checkstyle and cannot be kept. Preserving or strengthening summary validation would need a separate migration to `SummaryJavadoc`, with the existing violations fixed or suppressed.

vmassol:
> I tested `SummaryJavadoc`, which is the proper check for summary validation, but it reports many existing violations in XWiki: 5 in `xwiki-commons` and 428 in `xwiki-rendering`.

Thx for testing this. I don't see any param in `SummaryJavadoc` to just check that the first sentence ends properly, so I guess it's no longer possible. So if we want to use `SummaryJavadoc` we'll need to fix a lot of places (xwiki-platform is x100 compared to xwiki-rendering ...).

vmassol:
@gianmarcoschifone quick question (a bit out of scope). We want to allow using `{@returns ...}` (see https://forum.xwiki.org/t/add-a-global-search-box-to-documentation-home-page/18605) but when not having a `@return` tag I get:

```
JavadocMethod: @return tag should be present and have description.
```

I see that I could turn [allowMissingReturnTag](https://checkstyle.sourceforge.io/checks/javadoc/javadocmethod.html#allowMissingReturnTag) to `false` but that would fail to warn us if there's a method without a `@return` and without a `{@returns...}`.

What is the current status re the support of `{@returns...}`?

Thanks

gianmarcoschifone:
@vmassol 
I'm not sure if I understood correctly, but if you want to not have violation when there is either `@return` or `{@return}`, solution is to set [allowInlineReturn](https://checkstyle.sourceforge.io/checks/javadoc/javadocmethod.html#allowInlineReturn) to `true`, while [allowMissingReturnTag](https://checkstyle.sourceforge.io/checks/javadoc/javadocmethod.html#allowMissingReturnTag) needs to have default value (`false`), so no need to set it.

> Note that `{@returns}` is not correct Javadoc syntax, correct syntax is `{@return}`

vmassol:
@gianmarcoschifone thanks for the answer. I had missed [allowInlineReturn](https://checkstyle.sourceforge.io/checks/javadoc/javadocmethod.html#allowInlineReturn) inded. Works fine now.

PR #1831, opened 2026-07-11
[Misc] Fix SonarCloud S3077 bug in EmbeddableComponentManager

vmassol:
## Problem

SonarCloud reports a bug ([java:S3077](https://sonarcloud.io/project/issues?id=org.xwiki.commons%3Axwiki-commons&open=AZ9RL5Se1sKQ6MLE92AM)) on the `volatile R instance` field of `ComponentEntry` in `EmbeddableComponentManager`:

> Use a thread-safe type; adding "volatile" is not enough to make this field thread-safe.

Marking an object reference `volatile` only guarantees the visibility of the reference assignment, not of any compound operation, so Sonar recommends a thread-safe holder for such fields.

## Fix

Replace `private volatile R instance;` with `private final AtomicReference<R> instance = new AtomicReference<>();` and update all read/write sites (`.get()` / `.set()`). `AtomicReference` provides the same volatile read/write memory semantics that the existing double-checked-locking idiom relies on, so behaviour is unchanged while the warning is cleared.

## Verification

- `xwiki-commons-component-default` compiles and all 69 tests pass.
- 0 Checkstyle violations.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/main/java/org/xwiki/component/embed/EmbeddableComponentManager.java
+++ b/xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/main/java/org/xwiki/component/embed/EmbeddableComponentManager.java
@@ -31,6 +31,7 @@ import java.util.Objects;
 import java.util.ServiceLoader;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicReference;
 
 import jakarta.inject.Provider;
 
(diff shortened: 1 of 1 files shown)


tmortagne:
You might want to apply the same solution as on https://github.com/xwiki/xwiki-platform/commit/2c99b378a582f50d468fcd7a7ea57330b675a492 when fixing those double locking cases.

vmassol:
Thanks Thomas. I looked into applying `Suppliers.memoize()` here but it does not fit this particular case: unlike `XWiki#urlFactoryService` (a pure initialize-once-and-cache-forever singleton), `ComponentEntry#instance` has a mutable lifecycle. It gets reset to `null` on dispose and re-created on demand (e.g. `instance.set(null)` on disposal, `instance.set(createInstance(...))` on lazy creation, plus release/override handling). `Suppliers.memoize()` computes once and can never be reset or recomputed, so it cannot model a holder that is disposed and re-instantiated. That is why `AtomicReference` (a resettable thread-safe holder) is the right primitive here.

I will keep the `memoize()` pattern in mind for genuine lazy-init-once double-checked-locking cases though.

PR #1867, opened 2026-08-01
[Misc] Fix log message errors and Logging Best Practices violations

vmassol:
# Jira URL

N/A — `[Misc]` commit (log message wording and logging-convention fixes, no functional change).

# Changes

## Description

Audit of every logger call in the main sources against the [Logging Best Practices](https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/JavaCodeStyle/#HLoggingBestPractices) of the Java Code Style guide (395 calls), plus a read-through of the 304 literal log messages looking for wording errors.

**Message errors**

* Typos: `environmennt`, `remaning`, `any know descriptor`, `informations`, `Unsuported` (x2), `Uknown`, `fallbacking`, `will not contains`.
* `ChainingUberspector` logged *"Using SecureUberspect instead as a fallback"* — no such class; it is `SecureUberspector` (`LinkingUberspector` already had it right).
* `DefaultInstalledExtensionRepository` logged *"backward dependency index dependency [{}]"* — repeated word.

**Rule violations**

* *"Always pass a stack trace when logging errors"* — four `error()` calls had the caught exception in scope and dropped it entirely, logging no cause at all (`JarExtendedURLStreamHandler`, `ConverterRegistratorListener` x2, `MavenExtensionScanner`). They now pass it.
* *"Use the appropriate SLF4J signature"* — eight `logger.warn(String.format(...))` calls in the two legacy uberspectors converted to the parameterized form; their `e.getMessage()` replaced by `ExceptionUtils.getRootCauseMessage(e)`.
* *"Surround parameters with []"* — `does not implement {}` → `does not implement [{}]` in both uberspectors.
* *"Do not print a stack trace when you output a warning"* / *"error … always pass a stack trace"* — five `error()` calls passed only `getRootCauseMessage(e)`, which is the idiom the guide prescribes for `warn()`, and all of them recover (fallback or `return false`). Downgraded to `warn()`.
* *"info: only when it's absolutely necessary for the user to see the message"* — `DefaultDiffManager` reported a wrong-offsets anomaly at `info()`; downgraded to `warn()`.
* `DefaultDiffManager` also used a bare `logger` instead of `this.logger`.

## Clarifications

[...]

PR #1871, opened 2026-08-03
[Misc] Apply the logging best practices: root cause in warn(), no explicit toString(), static final Logger

vmassol:
The six-pass logging audit that ran over `xwiki-platform` (#6055 … #6079 there) had never been pointed at this repository. This is that scan, re-implemented as one parser covering all six passes' categories — placeholder arity and syntax, bracketing, `warn()` with a Throwable, concatenated / `String.format()` messages, `e.getMessage()`, calls with no message literal, `printStackTrace()`, `System.out`, redundant level guards, what a `catch` does with the exception it holds, and how the `Logger` field is acquired, named and modified. It resolves every `Logger`-typed identifier declared in a file rather than matching a receiver named `logger`, so a differently-named field is not invisible to it.

Over 1440 `src/main/java` files it reports 26 candidates. #1867 had already taken most of the `catch`-block ones; **12 are real and fixed here**, and the remaining 14 are non-violations, listed below.

### `warn(..., e.getMessage())` → `ExceptionUtils.getRootCauseMessage(e)` (5)

`EscapeTool.css` and four sites in `XMLUtils`. `getMessage()` on a wrapper says nothing about the underlying failure; the rule is to surface the root cause. The rendering gets *shorter* and gains the class name — `XMLUtilsTest` shows it best:

```
- Failed to apply XSLT transformation: [javax.xml.transform.TransformerException: com.sun.org.apache.xml.internal.utils.WrappedRuntimeException: Invalid url protocol: file]
+ Failed to apply XSLT transformation: [WrappedRuntimeException: Invalid url protocol: file]
```

### An explicit `.toString()` on a placeholder argument (4)

`ComponentAnnotationLoader`, `RepositoryUtils`, `DefaultJobProgress` (×2). SLF4J calls `toString()` on the argument itself, so writing it out is redundant, **eager** (paid even when the level is disabled) and **NPEs on a null** where SLF4J would have rendered `null`. That last point is not hypothetical in `DefaultJobProgress`: both calls are on the branch taken precisely because the step could not be matched from `source`. None of the four arguments is an array, so nothing gets spread as varargs.

### The rest (3)

- **`DefaultXMLParser`** discarded its `ConversionException` completely — the log said the conversion was unsupported but never why. `warn()` prints no trace, so the cause goes into a placeholder.
- **`EmbeddableComponentManager`** built a **new `Logger` per instance** (neither `static` nor `final`). Now `private static final LOGGER`, placed next to the `SHUTDOWN_LOGGER` that class already had; the field is private, so `TestComponentManager` and the other subclasses are unaffected.
- **`RuntimeUtils`** assembled its message with `String.format()` into a local, so the call site had no literal for a parser or a developer to find. The local stays (the return value needs it); only the log call is parameterized.

### The 14 candidates that are not violations

Recorded so a re-run recognises them rather than re-deciding:

- **`XWikiLogger` (extension-repository-maven) and `LogEvent`** pass a `Throwable` to `warn()`, but both are *adapters* replaying a `(message, throwable)` pair they were handed — the Maven/Aether logger bridge and a captured `LogEvent` being replayed. Same call as the platform's `RemoteListener`.
- **`FailingTestDebuggingTestExecutionListener`'s two concatenations** are in the `RuntimeUtils.run("docker …")` *command*, not in the message; the message is that command's output, which is the point of that listener.
- **`DefaultCoreExtensionScanner`'s `"Found the following JARs: {}"`** — the argument is a `Collection`, whose `toString()` already brackets its content.
- **Four `info()`/`debug()` calls in a `catch` that ignore the exception** (`LocalExtensionStorage`, `LogbackEventGenerator`, `BlobStoreMigrator` ×2) and **two `warn()` ones** (`LocalExtensionStorage`) — in each the catch *is* the documented outcome and the message already says what the exception would (`DirectoryNotEmptyException` → "was not empty … Keeping it", `FileNotFoundException` → "was not found", `BlobAlreadyExistsException` → "Detected existing migration marker … resuming").
- **`src/test/java` is a dry well**, as it was in the platform audit: 64 further candidates, all in two files — `logging-api`'s test `Utils`, a fixture that deliberately emits every SLF4J call shape (`"msg{}"`, markers, throwables) as *test data*, and `FindEntropyForSecureRandomProviderTest`, a diagnostic whose console output is its report.

### Verification

`mvn -B -ntp test` green from all eight touched modules — `velocity` (213), `xml` (289), `component-default` (69), `extension-api` (223), `job-api`, `job-default` (111), `filter-xml` (15), `tool-test-simple` (6) — and `mvn -B -ntp checkstyle:check -Pquality` green from each of them. Five test assertions were updated for the new rendering: four in `XMLUtilsTest` and one in `EscapeToolTest`. `DefaultJobProgressTest` needs no change — dropping `toString()` renders identically.

One line that a fix pushed to 123 characters was rewrapped.

### DIFF
--- a/xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/main/java/org/xwiki/component/annotation/ComponentAnnotationLoader.java
+++ b/xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/main/java/org/xwiki/component/annotation/ComponentAnnotationLoader.java
@@ -167,8 +167,7 @@ public class ComponentAnnotationLoader
             // Look for ComponentRole annotations and register one component per ComponentRole found
             Set<Type> componentRoleTypes = findComponentRoleTypes(componentClass);
             if (componentRoleTypes.isEmpty()) {
-                LOGGER.warn("The component implemented by class [{}] does not have any role type",
-                    componentClass.toString());
+                LOGGER.warn("The component implemented by class [{}] does not have any role type", componentClass);
             } else {
                 for (Type componentRoleType : componentRoleTypes) {
                     for (ComponentDescriptor<?> componentDescriptor : this.factory.createComponentDescriptors(
--- a/xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/main/java/org/xwiki/component/embed/EmbeddableComponentManager.java
+++ b/xwiki-commons-core/xwiki-commons-component/xwiki-commons-component-default/src/main/java/org/xwiki/component/embed/EmbeddableComponentManager.java
@@ -71,6 +71,8 @@ import org.xwiki.component.util.ReflectionUtils;
 @SuppressWarnings("checkstyle:ClassFanOutComplexity")
 public class EmbeddableComponentManager implements NamespacedComponentManager, Disposable
 {
+    private static final Logger LOGGER = LoggerFactory.getLogger(EmbeddableComponentManager.class);
+
     /**
      * Logger to use to log shutdown information (opposite of initialization).
      */
@@ -218,7 +220,6 @@ public class EmbeddableComponentManager implements NamespacedComponentManager, D
 
     private ConcurrentMap<Type, ComponentEntries> componentEntries = new ConcurrentHashMap<>();
 
-    private Logger logger = LoggerFactory.getLogger(EmbeddableComponentManager.class);
 
     /**
      * Finds all lifecycle handlers to use when instantiating a Component.
@@ -280,7 +281,7 @@ public class EmbeddableComponentManager implements NamespacedComponentManager, D
             }
         } catch (ComponentLookupException e) {
             // Should never happen
-            this.logger.error("Failed to lookup ComponentManagerInitializer components", e);
+            LOGGER.error("Failed to lookup ComponentManagerInitializer components", e);
         }
     }
 
@@ -440,7 +441,7 @@ public class EmbeddableComponentManager implements NamespacedComponentManager, D
                 throw new ComponentLookupException("Failed to lookup component with type [%s] and hint [%s]"
                     .formatted(roleEntry.descriptor.getRoleType(), roleEntry.descriptor.getRoleHint()), e);
             } else {
-                this.logger.error("Failed to lookup component with type [{}] and hint [{}]",
+                LOGGER.error("Failed to lookup component with type [{}] and hint [{}]",
                     roleEntry.descriptor.getRoleType(), roleEntry.descriptor.getRoleHint(), e);
             }
         }
@@ -882,7 +883,7 @@ public class EmbeddableComponentManager implements NamespacedComponentManager, D
         try {
             removeComponent(role, hint);
         } catch (Exception e) {
-            this.logger.warn("Instance released but disposal failed. Some resources may not have been released: [{}]",
+            LOGGER.warn("Instance released but disposal failed. Some resources may not have been released: [{}]",
                 ExceptionUtils.getRootCauseMessage(e));
         }
     }
@@ -971,7 +972,7 @@ public class EmbeddableComponentManager implements NamespacedComponentManager, D
                         disposable.dispose();
                         SHUTDOWN_LOGGER.debug("Component [{}] has been disposed", instance.getClass().getName());
                     } catch (ComponentLifecycleException e) {
-                        this.logger.error("Failed to dispose component with role type [{}] and role hint [{}]",
+                        LOGGER.error("Failed to dispose component with role type [{}] and role hint [{}]",
                             componentEntry.descriptor.getRoleType(), componentEntry.descriptor.getRoleHint(), e);
                     }
                 }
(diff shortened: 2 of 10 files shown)


tmortagne:
I'm afraid "no explicit toString()" is not a good rule as our log is often serialized. So many of the toString you "fixed" in the code were here on purpose (to serialize a String instead of something much bigger we don't really need).

vmassol:
> I'm afraid "no explicit toString()" is not a good rule as our log is often serialized. So many of the toString you "fixed" in the code were here on purpose (to serialize a String instead of something much bigger we don't really need).

Can you explain a bit more as I don't understand. Why would we want to call `toString()` on any log statement (even if the log is not active)? And what's the difference of calling `toString()` early, vs letting SLF4J call it? I don't understand the argument of "something much bigger". Thx

tmortagne:
> And what's the difference of calling toString() early, vs letting SLF4J call it? I don't understand the argument of "something much bigger".

SLF4J does not call toString.

You only have the `catalina.out` use case in mind (where logback serialize the log event as a big string and just use toString for each parameter, because it's in a context where there is no concept of syntax). But it's not how job log storage works: the log event is serialized in a way in which it can be unserialized, because parameters have special displayer depending on their type (links for entity references and extension id, etc.) in the log displayer.

vmassol:
You're right, and I checked the code path to make sure I understood why rather than just taking your word for it. Recording it here since the reason isn't obvious from a call site.

`AbstractJobStatus` pushes a log listener, so any `warn`/`error` executed on a job thread is captured whatever class it comes from. The captured `LogEvent` keeps the raw `Object[]`, and `SafeMessageConverter.marshal()` then does:

```java
if (isSerializable(argument)) {
    writeCompleteItem(argument, context, writer);   // full object graph
} else {
    writeCompleteItem(argument.toString(), context, writer);
}
```

And `XStreamUtils.isSerializable()` ends with `return true` — anything not a `@Component`/`Logger`/`Provider`/stream is serialized completely. So dropping an explicit `toString()` isn't a no-op: it changes what goes into the job log file, and on the way back `SafeArrayConverter.readBareItem()` swallows read failures and yields `null`, so an argument whose class can no longer be resolved is lost entirely where the string would have survived. Your displayer point has teeth beyond display too — `Importer` casts `log.getArgumentArray()[0]` to `EntityReference`.

Two of my three arguments in the description don't survive that: "eager" is irrelevant since all these sites are `warn`/`error` (always enabled), and the NPE concern is fixable with `String.valueOf(x)`, which keeps the string eager *and* null-safe.

So I'll revert, with `String.valueOf()` and a comment explaining the mechanism so the next pass doesn't re-decide it:

- `DefaultJobProgress` ×2 — arbitrary `Object source`, and this code runs inside a job by definition. Clearest case.
- `ComponentAnnotationLoader` — `getName()`; it warns about classes from extension jars, i.e. the ones most likely unresolvable later.
- `RepositoryUtils` I'd leave as merged unless you disagree: `ExtensionRepositoryDescriptor` is a commons type that's always resolvable, and its `toString()` already prints every field, so there's no size win and the typed form is the more useful one.

The same pass hit xwiki-platform, where ~6 sites need the same treatment — worst is `XWikiHibernateBaseStore` now passing a live Hibernate `Session` (which extends `Serializable`, so XStream will happily walk it), plus `MovedAttachmentListener` passing a `MoveAttachmentRequest` from inside the move job itself, the two Solr listeners passing arbitrary `source`, and two `StringBuilder`s. Separate PR.

One catch: `java:S2629` ("logging arguments should not require evaluation") is what closed two of these in SonarCloud, so part of the revert re-opens it. Reading `LazyArgEvaluationCheck`, it only examines arguments whose static type is `String`, and within those it exempts no-arg `get*`/`is*` calls and skips log calls inside a `catch` block entirely. So `String.valueOf(source)` in `DefaultJobProgress` re-triggers it — I'll add `@SuppressWarnings("java:S2629")` there with the explanatory comment, following the `java:S2789` precedent in `ServletEnvironment`. `ComponentAnnotationLoader` needs no suppression since `getName()` hits the getter exemption, and `RepositoryUtils` sits in a `catch`, so the rule never had an opinion on it either way.

Which brings me to a follow-up I'd like your opinion on, *after* the revert, not now. The real problem is that `isSerializable()` defaults to `true` for log arguments — for an unknown type that default is backwards. Would you be OK with inverting it in the log-argument path only: `SafeMessageConverter` keeps the object typed when there's a positive signal (`isSafeType()`, or an explicit `@Serializable(true)`) and stringifies everything else, with the types the displayers and `Importer` rely on annotated `@Serializable(true)`? It stays scoped to log arguments so job status/request serialization is untouched, it handles arbitrary `Object` sources that no annotation can reach, and it would let us delete all these call-site `toString()`s and their comments. The cost is surveying what currently relies on typed log arguments — you'd know that far better than I do, which is why I'm asking rather than proposing a patch.

tmortagne:
> The real problem is that isSerializable() defaults to true for log arguments

I'm afraid it's not that simple. Any extension might rely on that for various reasons, so it's not that easy to know.
It would also not cover any argument we don't want to serialize but which happen to be `java.io.Serializable` for other needs, so it's not enough to cover all use cases.

We eventually need to work on making log argument displayers more component based (the log display is quite hackish right now on this area). When someone finally start working on that, we could also discuss various way to declare classes for which log argument serialization is allowed for various reasons (those having a corresponding displayer, those with some explicit annotation, those listed by some components because you don't have control over theses classes, etc.)

PR #1963, opened 2026-09-10
XCOMMONS-2092: Possible race condition in ReadWriteSemaphore

surli:
## Jira URL

https://jira.xwiki.org/browse/XCOMMONS-2092

### Description

`DefaultJobExecutorTest#matchingGroupPathAreBlockedPoolMultiSizeParentFirst` occasionally failed on CI with:
```
org.opentest4j.AssertionFailedError: expected: <null> but was: <WAITING>
	at org.xwiki.job.internal.DefaultJobExecutorTest.matchingGroupPathAreBlockedPoolMultiSizeParentFirst(DefaultJobExecutorTest.java:273)
```

This went through three iterations, kept as three commits, each building on what the previous one got wrong or left incomplete:

1. **First commit** — initial read of the failure: a timing gap between submitting a writer and submitting a reader that must queue behind it. Fixed with a `Thread.sleep(WAIT_VALUE)`.

2. **Second commit** — the actual root cause found after [a JIRA comment](https://jira.xwiki.org/browse/XCOMMONS-2092?focusedId=122511&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-122511) pointed at `ReadWriteSemaphore` itself: `lockWrite()`/`unlockWrite()`/`lockRead()`/`unlockRead()` mutated `writeCounter`/`readCounter` with `incrementAndGet()`/`decrementAndGet()`, then decided how many semaphore permits to acquire/release from a **separate**, later `.get()` — a TOCTOU race that could permanently lose a permit. Fixed by making the counter mutation and the resulting permit-count decision atomic (one `synchronized` block), and removed the first commit's `Thread.sleep` as apparently no longer needed.

3. **Third commit (this one)** — the second commit's fix, while it did close the TOCTOU race, kept the same fragile design of encoding read/write exclusion as a *number of permits* computed from a counter snapshot. Validating it at scale (`@RepeatedTest` in the tens of thousands, per a JIRA comment on reproducing this locally) surfaced that this whole approach is fragile beyond just the atomicity:
   - `matchingGroupPathAreBlockedPoolMultiSizeChildrenFirst` failed consistently (roughly half the time) once `ReadWriteSemaphore` was rewritten to fix the TOCTOU race the naive way — it turns out readers and writers actually **share one pool of slots** (that test lets a writer and a reader be concurrently active once the pool size allows it), a sizing rule the permit-count scheme got subtly wrong.
   - Once that was fixed, the *same* test then failed intermittently because plain `wait()`/`notifyAll()` gives no FIFO guarantee: a slot freed up for the earliest-still-waiting writer could be taken by a later one instead.
   - With both fixed, `matchingGroupPathAreBlockedPoolMultiSizeParentFirst` still flickered at roughly the same rate it originally did. A captured thread dump confirmed why: a later-submitted reader's worker thread can still reach `lockRead()` before an earlier-submitted writer's own thread reaches `lockWrite()` — a property of when the surrounding thread pool dispatches each one, which nothing inside `ReadWriteSemaphore` can control. The first commit's `Thread.sleep` closed exactly this window; it is restored here (as a targeted, commented wait at the one place it's needed, not a blanket reversion).

4. **Fourth commit** — SonarCloud reported `java:S5961` (28 assertions, threshold 25) on `matchingGroupPathAreBlockedPoolMultiSizeParentFirst`. It was first suppressed; this commit cleans the test up for real instead. SonarJava counts assertions reached through private helpers, so 10 of the 28 were the `fail()` inside `waitJobState()`, and 12 of the 18 remaining direct assertions only restated the state the `waitJobXxx()` call right above them had already established. Those restatements are dropped in all three `matchingGroupPath*` tests, the repeated `GroupedJobInitializer` stubbing is extracted into `mockPool(poolSize, path...)` / `mockAllPools(poolSize)`, and a duplicated assertion on `job1` (meant to be on `job12`, and already covered) is removed. Splitting the methods instead was rejected: each is a single choreography over live threads whose assertions are checkpoints on one timeline, so every split test would have to replay all the steps before its own.

### Clarifications

`ReadWriteSemaphore` is rewritten to a plain monitor (`synchronized`/`wait`/`notifyAll`) checking `readCounter`/`writeCounter`/`activeWriters` directly, instead of translating the same invariant into `Semaphore` permit counts. This removes the whole *class* of "permits taken and given back don't match" bugs by construction — there is no derived permit count to keep resynchronized with reality any more. Waiting writers are served in `lockWrite()` call order via an explicit ticket, to guarantee the FIFO property `wait()`/`notifyAll()` doesn't provide on its own.

## Executed Tests

- `mvn clean install -B -ntp -pl xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default -Plegacy` — full module build (125 tests, Checkstyle, license headers, Revapi) green
- New `ReadWriteSemaphoreTest`: a concurrent stress test that reliably deadlocks the pre-atomicity-fix (first-commit) code within a few hundred operations, and passes cleanly against the current code
- After the fourth commit: `mvn clean install -B -ntp -pl xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default -Plegacy` green again (125 tests, Checkstyle), plus 20 consecutive standalone runs of `DefaultJobExecutorTest` with 0 failures
- All three `DefaultJobExecutorTest` methods, each run standalone in its own fresh JVM via `@RepeatedTest`: `matchingGroupPathAreBlocked` 5000/5000, `matchingGroupPathAreBlockedPoolMultiSizeParentFirst` 5000/5000, `matchingGroupPathAreBlockedPoolMultiSizeChildrenFirst` 1500/1500 — 0 failures across 11,500 repetitions

## Expected merging strategy

Prefers squash: Yes

### DIFF
--- a/xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/test/java/org/xwiki/job/internal/DefaultJobExecutorTest.java
+++ b/xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/test/java/org/xwiki/job/internal/DefaultJobExecutorTest.java
@@ -182,7 +182,7 @@ class DefaultJobExecutorTest
     }
 
     @Test
-    void matchingGroupPathAreBlockedPoolMultiSizeParentFirst()
+    void matchingGroupPathAreBlockedPoolMultiSizeParentFirst() throws InterruptedException
     {
         // Check the following setup:
         // Pool A of size 1 with 2 jobs (A1, A2)
@@ -263,6 +263,10 @@ class DefaultJobExecutorTest
         jobAB1.unlock();
 
         waitJobFinished(jobAB1);
+
+        // We wait a bit to ensure A2 actually requested the lock before starting AB3.
+        // FIXME: We cannot use waitJobWaiting since the job itself is not started yet (blocked by previous jobs)
+        Thread.sleep(WAIT_VALUE);
 
         // Start AB3 only now to be sure it does not take the lock before A2.
         this.executor.execute(jobAB3);
--- a/xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/test/java/org/xwiki/job/internal/DefaultJobExecutorTest.java
+++ b/xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/test/java/org/xwiki/job/internal/DefaultJobExecutorTest.java
@@ -182,7 +182,7 @@ class DefaultJobExecutorTest
     }
 
     @Test
-    void matchingGroupPathAreBlockedPoolMultiSizeParentFirst() throws InterruptedException
+    void matchingGroupPathAreBlockedPoolMultiSizeParentFirst()
     {
         // Check the following setup:
         // Pool A of size 1 with 2 jobs (A1, A2)
@@ -263,10 +263,6 @@ class DefaultJobExecutorTest
         jobAB1.unlock();
 
         waitJobFinished(jobAB1);
-
-        // We wait a bit to ensure A2 actually requested the lock before starting AB3.
-        // FIXME: We cannot use waitJobWaiting since the job itself is not started yet (blocked by previous jobs)
-        Thread.sleep(WAIT_VALUE);
 
         // Start AB3 only now to be sure it does not take the lock before A2.
         this.executor.execute(jobAB3);
--- a/xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/ReadWriteSemaphore.java
+++ b/xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/main/java/org/xwiki/job/internal/ReadWriteSemaphore.java
@@ -20,7 +20,6 @@
 package org.xwiki.job.internal;
 
 import java.util.concurrent.Semaphore;
-import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * A specific concurrency implementation for managing Semaphore with Read/Write lock capabilities.
@@ -32,8 +31,19 @@ import java.util.concurrent.atomic.AtomicInteger;
  */
 public class ReadWriteSemaphore
 {
-    private final AtomicInteger readCounter;
-    private final AtomicInteger writeCounter;
+    /**
+     * Protects {@link #readCounter} and {@link #writeCounter}: reading a counter and deciding, from that read, how
+     * many permits to acquire or release from {@link #semaphore} must happen as a single atomic step, otherwise
+     * concurrent callers can compute an inconsistent number of permits and permanently lose some, deadlocking every
+     * subsequent caller. The lock is only ever held for that bookkeeping, never while blocked in the semaphore
+     * itself, so a thread waiting for a permit never holds a lock another thread needs to release one.
+     */
+    private final Object countersLock = new Object();
+
+    private int readCounter;
+
+    private int writeCounter;
+
     private final Semaphore semaphore;
 
     /**
@@ -43,8 +53,6 @@ public class ReadWriteSemaphore
     public ReadWriteSemaphore(int poolSize)
     {
         this.semaphore = new Semaphore(poolSize, true);
-        this.readCounter = new AtomicInteger(0);
-        this.writeCounter = new AtomicInteger(0);
     }
 
     /**
@@ -53,13 +61,15 @@ public class ReadWriteSemaphore
      */
     public void lockWrite()
     {
-        this.writeCounter.incrementAndGet();
+        int permits;
 
-        if (this.writeCounter.get() == 1) {
-            this.semaphore.acquireUninterruptibly(this.readCounter.get() + 1);
-        } else {
-            this.semaphore.acquireUninterruptibly();
+        synchronized (this.countersLock) {
+            this.writeCounter++;
+
+            permits = this.writeCounter == 1 ? this.readCounter + 1 : 1;
         }
+
+        this.semaphore.acquireUninterruptibly(permits);
     }
 
     /**
@@ -68,13 +78,15 @@ public class ReadWriteSemaphore
      */
     public void unlockWrite()
     {
-        this.writeCounter.decrementAndGet();
+        int permits;
 
-        if (this.writeCounter.get() == 0) {
-            this.semaphore.release(this.readCounter.get() + 1);
-        } else {
-            this.semaphore.release();
+        synchronized (this.countersLock) {
+            this.writeCounter--;
+
+            permits = this.writeCounter == 0 ? this.readCounter + 1 : 1;
         }
+
+        this.semaphore.release(permits);
     }
 
     /**
@@ -83,9 +95,15 @@ public class ReadWriteSemaphore
      */
     public void lockRead()
     {
-        this.readCounter.incrementAndGet();
+        boolean needPermit;
 
-        if (this.writeCounter.get() > 0) {
+        synchronized (this.countersLock) {
+            this.readCounter++;
+
+            needPermit = this.writeCounter > 0;
+        }
+
+        if (needPermit) {
             this.semaphore.acquireUninterruptibly();
         }
     }
@@ -96,9 +114,15 @@ public class ReadWriteSemaphore
      */
     public void unlockRead()
     {
-        this.readCounter.decrementAndGet();
+        boolean hadPermit;
 
-        if (this.writeCounter.get() > 0) {
+        synchronized (this.countersLock) {
+            this.readCounter--;
+
+            hadPermit = this.writeCounter > 0;
+        }
+
+        if (hadPermit) {
             this.semaphore.release();
         }
     }


xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/test/java/org/xwiki/job/internal/DefaultJobExecutorTest.java
```diff
@@ -264,6 +264,10 @@ void matchingGroupPathAreBlockedPoolMultiSizeParentFirst()
 
         waitJobFinished(jobAB1);
 
+        // We wait a bit to ensure A2 actually requested the lock before starting AB3.
+        // FIXME: We cannot use waitJobWaiting since the job itself is not started yet (blocked by previous jobs)
+        Thread.sleep(WAIT_VALUE);
```
manuelleduc: See my last comment on https://jira.xwiki.org/browse/XCOMMONS-2092
I think this would hide a rare but real product issue.

manuelleduc:
I guess the issue title is not accurate anymore :)
I don't know enough to evaluate the proposed fix, possibly more something for @michitux or @tmortagne.

surli:
> I guess the issue title is not accurate anymore :)

Well it's still the title of the original ticket, I guess I should maybe create another ticket and close the original one as solved by?

tmortagne:
The title of the issue is definitely not accurate anymore since it's fixing an actual bug, and it's not a test thing. You could repurpose the issue IMO, no need to create a new one.

tmortagne:
>  I don't know enough to evaluate the proposed fix, possibly more something for @michitux or @tmortagne.

I usually try to avoid implementing stuff like this myself. Not that confident in my ability to spot problems like this (I reviewed that buggy version back then after all....). I remember we could not find something to reuse back then. I asked Claude to look at Apache Commons and Guava alternatives again, but it did not really found anything interresting.

surli:
@tmortagne / @manuelleduc yeah actually I was wondering if we really wanted to backport that, given the fact we're not so sure about this fix. When reading the code it feels logical, but as Thomas said the previous code also felt correct. 
That being said, and as indicated in the description, I ran 500 occurrences of the unit tests without any failure, which gives some guarantees IMO.

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/test/java/org/xwiki/job/internal/DefaultJobExecutorTest.java
```diff
@@ -152,15 +150,11 @@ void matchingGroupPathAreBlocked()
         waitJobWaiting(jobAB);
 
         assertSame(State.FINISHED, jobA.getStatus().getState());
-        assertSame(State.WAITING, jobAB.getStatus().getState());
 
         // Next job
         jobAB.unlock();
         waitJobFinished(jobAB);
 
-        assertSame(State.FINISHED, jobA.getStatus().getState());
```
surli, line 161: The wait above only wait on jobAB this assertion is still needed.

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/test/java/org/xwiki/job/internal/DefaultJobExecutorTest.java
```diff
@@ -171,21 +165,14 @@ void matchingGroupPathAreBlocked()
...
 
         assertSame(State.FINISHED, job12.getStatus().getState());
-        assertSame(State.WAITING, job1.getStatus().getState());
 
         // Next job
         job1.unlock();
         waitJobFinished(job1);
-
-        assertSame(State.FINISHED, job1.getStatus().getState());
-        assertSame(State.FINISHED, job1.getStatus().getState());
```
surli, line 181: I'm not a big fan of remove those assert as it makes the test easier to read to see the status of each job IMO

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/test/java/org/xwiki/job/internal/DefaultJobExecutorTest.java
```diff
@@ -280,7 +253,6 @@ void matchingGroupPathAreBlockedPoolMultiSizeParentFirst() throws InterruptedExc
         this.executor.execute(jobAB3);
 
         // AB3 cannot start yet even if the pool is of 2 because A2 requested for the lock.
-        assertSame(State.FINISHED, jobAB1.getStatus().getState());
```
surli, line 271: There's no wait for this one, so we should still have this assert

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/test/java/org/xwiki/job/internal/DefaultJobExecutorTest.java
vmassol: You're right that `waitJobFinished(jobAB)` says nothing about `jobA`. My reason was a different one: `assertSame(FINISHED, jobA)` four lines above already covers it, `FINISHED` is terminal (`AbstractJob` sets it once in its `finally`), and the group lock is only released afterwards in `DefaultJobExecutor#afterExecute`, so that earlier assertion isn't racy either.

That said, I've restored it — the count is now fixed at its source instead. See the general comment below.

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/test/java/org/xwiki/job/internal/DefaultJobExecutorTest.java
vmassol: Agreed, and I shouldn't have touched this method at all: it counts 12 assertions against a threshold of 25, so it was never flagged and never could have been. All restored.

The one thing I kept is the last assertion, which was on `job1` twice — I read it as meaning `job12` and `job1`, which is what it is now.

xwiki-commons-core/xwiki-commons-job/xwiki-commons-job-default/src/test/java/org/xwiki/job/internal/DefaultJobExecutorTest.java
vmassol: There is a wait for this one: `waitJobFinished(jobAB1)` twelve lines above, just before the `Thread.sleep` block — easy to miss with that comment sitting in between.

Restored anyway, along with all the others.

vmassol:
@surli thanks — all three comments applied in 436b01d4: every assertion is back, the three methods are identical to before my earlier commit in that respect. I took a different route to clear `java:S5961`.

**What was actually inflating the count.** Assertion-aware analysis resolves helper methods transitively (`TooManyAssertionsCheck#isAssertion` ends with `|| !collectAssertionsInMethod(methodSymbol).isEmpty()`), and `waitJobState()` reported its timeout through `fail()`. So every `waitJobWaiting()` / `waitJobFinished()` call counted as an assertion of the calling test — 10 of the 28 reported on `matchingGroupPathAreBlockedPoolMultiSizeParentFirst`.

That means each checkpoint of these tests was counted twice: once for the barrier that drives the choreography to its next step, and once for the assertion that states what is expected there. Your three comments are exactly the argument that those are two different things — if the wait were the assertion, the `assertSame` right after it would be noise and you'd have let me delete it.

**The change.** `waitJobState()` now throws `AssertionError` for its timeout instead of calling `fail()`. `Assertions.fail()` is itself a throw of an `AssertionError` subclass, so behaviour and Surefire reporting are unchanged, but the count now equals the number of expectations each test actually states:

| method | before | after |
|---|---|---|
| `…ParentFirst` | 28 | 18 |
| `…ChildrenFirst` | 29 | 17 |
| `matchingGroupPathAreBlocked` | 18 | 12 |

So the `@SuppressWarnings("java:S5961")` is gone, and `…ChildrenFirst` — which was over the threshold too, just not as new code — comes down as well.

There's a second reason beyond the count: `AssertionsInTestsCheck` (S2699, "tests should include assertions") resolves helpers the same way, via `isLocalMethodWithAssertion`. With `fail()` in a shared helper, a test whose only assertion-like call is a wait would be reported as having assertions. That's the dangerous direction, and it no longer happens. The reasoning is recorded as Javadoc on `waitJobState()` so it survives.

Also in this commit: the interrupt flag is restored and the `InterruptedException` kept as the cause instead of both being swallowed; the repeated `GroupedJobInitializer` stubbing is extracted into `mockPool(poolSize, path...)` / `mockAllPools(poolSize)`; and two comments mangled by an `A` → `AB` search and replace are fixed ("taken into ABcount", "they are ABtually after").

Verified locally: full module build green (125 tests, Checkstyle), plus 20 consecutive standalone runs of `DefaultJobExecutorTest` with no failures. The counts above are derived from the analyzer's own source rather than from a run — SonarCloud's analysis of this PR is the real confirmation.

vmassol:
SonarCloud has now analysed `436b01d4` (check run `SonarCloud Code Analysis`, 14:36:45Z, success): `java:S5961` is **closed** on this PR, no open issues remain, quality gate OK. So the counts above are confirmed rather than predicted — the three methods keep every assertion and the rule no longer fires.

